PR #24, opened 2024-09-26
Update jakanddaxter.toml to v0.1.2

massimilianodelliubaldini:


massimilianodelliubaldini:
Hmmm, Eije the bot that generates the Review Changes page doesn't seem to be able to differentiate Renamed files from brand new files, they all come in as Renamed. See `jakanddaxter/test/test_orbsanity.py`.

Eijebong:
That's just a visual bug because they get "renamed" from `/dev/null` and the library I use to generate the web page doesn't know how to handle it 🤷

PR #27, opened 2024-09-29
Add Manual for FF14

cloudofdarkhole6:
told a friend about a manual for FF14, so i guess i have to add this now

### DIFF
--- a/index/manual_ffxiv_silasary.toml
+++ b/index/manual_ffxiv_silasary.toml
@@ -1,7 +1,6 @@
 display_name = "Manual: Final Fantasy XIV"
 name = "Manual_FFXIV_Silasary"
 home = "https://discord.com/channels/1097532591650910289/1187584337877684294"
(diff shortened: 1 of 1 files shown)


Eijebong:
`Data.py:147`

This code can't make it in for 2 reasons.

1) Outside network access at runtime.
2) The yaml validator runs on a read only filesystem. This will just crash if we ever take that branch.

Now if this branch is never taken except when the apworld gets updated we could do with it but unless I know that, this is not happening.

```py
    if missing_fatesanity_zones:
        # This is hacky, but it lets me slowly scrape the wiki for FATEs without abusing the API
        for key in list(missing_fatesanity_zones.keys()):
            from . import wiki_scraper
            import os
            additional = wiki_scraper.find_fates(key)
            fates_path = os.path.join(os.path.dirname(__file__), 'fates.csv')
            with open(fates_path, 'a', newline='') as csvfile:
                writer = csv.writer(csvfile, delimiter=',', quotechar='|', quoting=csv.QUOTE_MINIMAL)
                for line in additional:
                    writer.writerow(line.split(','))
```

PR #29, opened 2024-10-06
Update Stardew Valley to 6.5.2

gerbiljames:
The .apworld is still not available through a public URL, so I've updated the local copy of it


Eijebong:
I don't think we can do that yet.

```diff
-        self.multiworld.state.collect(summer, event=False)
+        self.multiworld.state.collect(summer, prevent_sweep=False)
```

That is a 0.5.1 only arg rename from https://github.com/ArchipelagoMW/Archipelago/commit/96d48a923a34a1cd4c5f64a5f2e12acc657dd041#diff-7fd3235b44ff24c46065aa28ed907404938d667d1a4a5658c9dff338bbc55c26R866

I'll come back to this once 0.5.1 drops, hopefully it won't be needed anymore but we'll see.

gerbiljames:
Ah, just realised I did my testing on `main`

Hopefully they'll just update it directly since its supported

Eijebong:
So something I just realized. Those 0.5.1 changes are in tests only. And it looks like the validator goes through generation just fine. So i'll just merge this and hope nothing blows up

PR #31, opened 2024-10-06
Update Tunic to 4.0.1

gerbiljames:
I promise not to use the grass randomizer


Eijebong:
Yikes... Because it goes from a supported game to an unsupported one, that apdiff is completely messed up. I'll try fixing that tomorrow and then I'll have another look at this.

PR #45, opened 2024-10-29
Update Outer Wilds to 0.3.6

gerbiljames:


Eijebong:
Can you keep the old versions? Due to recent changes on the lobby, rooms can now specify apworld versions and while I doubt that anyone has used it yet or that someone would want to run an old version of outer wilds, we should keep a few "old" versions in the index from now on.

gerbiljames:
Done! Being able to pick apworld versions when creating a room is a really useful feature

Eijebong:
Had one reservation about the pickled files in here being unpickled with `pickle.load`. Contacted the author about it and manually checked that the ones in the apworld match the jsonc.

PR #66, opened 2024-11-05
Update megamix to 1.1.4-beta

RePod:
With argv patch.

The .idea folder is there because it's in the official release.


Eijebong:
Let's see if the validator agrees :p

Eijebong:
Not that it matters that much until that `ast` issue is resolved but I'm curious to know if the validator agrees now

PR #67, opened 2024-11-06
Update powerwash simulator manual

Eijebong:


Eijebong:
The world doesn't load when asserts are on because of them adding a `random` option value which archipelago doesn't like. Bug reported upstream [here](https://discord.com/channels/1097532591650910289/1181412855187243079/1303783830141796394)

PR #76, opened 2024-11-22
Update Anodyne to 0.1.7

Mysteryem:
0.1.6 and 0.1.7 should just be logic fixes backported from the development branch.


Eijebong:
I guess for something like this where the apworld actually respects semver, it makes sense to not keep the old version around yeah.

PR #79, opened 2024-11-29
Add Manual for Arcaea

Eijebong:


Eijebong:
This is blocked by the direct usage of `random` instead of `world.random` in `before_create_items_filler`

PR #123, opened 2025-01-09
Add no100f

Eijebong:


Eijebong:
This one is going to be a pain, it assumes one got DME installed. Probably gotta start installing it in the checker as it becomes more common.

I don't know about that .NET package either, if it tries to have a runtime at verification time we're gonna have a problem

Eijebong:
Asked about the pythonnet stuff as it seems to be some artifacts from the original BFBB import. https://discord.com/channels/731205301247803413/1188152412498829352/1326976681788964980

Eijebong:
It also occurred to me that the license might be an issue since it got re-licensed from GPL to MIT but I cannot find a clear approval for that anywhere. I contacted the author of said GPL code to try and clarify things.

Eijebong:
Got confirmation that it was all good. So the only blocking thing now is me finding motivation to repack this with pythonnet and friends removed

PR #127, opened 2025-01-11
Add empty manual and pokemon platinum

Eijebong:


silasary:
Technically, the template is a fully functional and playable Manual for Marvel Vs Capcom 3. 

I've only seen it played once, but I don't see a reason to blacklist it

PR #142, opened 2025-01-17
Add wl4 and autopelago

Eijebong:


Eijebong:
/taskcluster r+

wl4 has some direct random usage with music shuffle but it's cosmetic only

PR #150, opened 2025-01-18
Updates LADX Beta to 11.0

palex00:


Eijebong:
Can you keep the previous version? This way if we need to roll it back it's a matter of me changing a value in a dropdown. (just dupe the line)

PR #156, opened 2025-01-20
Add new game Ratchet & Clank 2

cloudofdarkhole6:
apworld gets shipped inside of a zip file, so had to save it local

### DIFF
--- a/index/rac2.toml
+++ b/index/rac2.toml
@@ -1,0 +1,6 @@
+name = "Ratchet & Clank 2"
+home = "https://discord.com/channels/731205301247803413/1325015730218860554"
+default_url = "https://github.com/evilwb/APRac2/releases/download/v{{version}}/rac2.apworld"
+
+[versions]
+"0.4.2" = {}


Mysteryem:
Note that this world can create unloadable multidata with TUNIC currently, and any other worlds trying to put RaC2 location names directly into slot data https://github.com/evilwb/APRac2/issues/1

Such a multiworld will generate fine, but will then be impossible to upload onto archipelago.gg.

There's a chance some games might also write RaC2 location names to their patch files and result in a broken patch.

RaC2's item names previously had a similar issue with The Wind Waker's patch files, which has since been fixed.

RaC2's region names also look like they would cause similar issues if a world were to try to put them into multidata.

cloudofdarkhole6:
New version released that fixes incompatibilities caused by StrEnum. So far no reports came in about anything broken and i also genned + hosted a combo together with Tunic which was where the issue original got found, generation + hosting on local worked.
Apworld is also now directly in release and not packed inside a .zip so i removed it from the commit

Eijebong:
> FileNotFoundError: [Errno 2] No such file or directory: '/builds/worker/.cache'

Love to hit the same bugs that I've seen at work...

PR #157, opened 2025-01-20
Add new game Golden Sun The Lost Age

cloudofdarkhole6:
Rom.py has a network request (POST to gs2randomiser.com), but it doesn't get called on generation, only during the patching


Eijebong:
#116

PR #193, opened 2025-02-03

RePod:
[The sequel.](https://store.steampowered.com/app/1057090/) Drafted to allow world review and testing to happen.
> "I will probably make a proper release with more fixes in a few days."

\- Dev

Let me know if the final PR should include 0.0.2 or the "proper release".

### DIFF
--- a/index/ori_wotw.toml
+++ b/index/ori_wotw.toml
@@ -3,4 +3,4 @@ home = "https://discord.com/channels/731205301247803413/1272952565843103765"
 default_url = "https://github.com/Satisha10/APwotw_release/releases/download/v{{version}}/ori_wotw.apworld"
 
 [versions]
-"0.3.1" = {}
+"0.3.3" = {}


Mysteryem:
This world is missing indirect conditions for every entrance rule that uses `can_buy_map` or `can_keystones` (100s of rules) and does not use `explicit_indirect_conditions = False`, so this can result in nondeterministic generation, or FillErrors from logic not re-checking entrances that might have become accessible because a region has just become accessible.

All the connections to the `Victory` region are also missing indirect conditions for the region access used in those rules.

See https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/world%20api.md#an-important-note-on-entrance-access-rules and https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/apworld_dev_faq.md#i-learned-about-indirect-conditions-in-the-world-api-document-but-i-want-to-know-more-what-are-they-and-why-are-they-necessary

Eijebong:
Indirect condition issue reported here: https://discord.com/channels/731205301247803413/1272952565843103765/1340369222047498330

Mysteryem:
The world is now using `explicit_indirect_conditions = False`. There is a TODO comment, so it's probably only a temporary measure until they manage to sort out indirect conditions.

Eijebong:
Does this still need to be a draft?

Eijebong:
Test failure reported here https://discord.com/channels/731205301247803413/1272952565843103765/1352717636957372477

Fuzzing failure rate is catastrophic (1/3), reported here https://discord.com/channels/731205301247803413/1272952565843103765/1352724423999754353

Eijebong:
Tests are passing but fuzzing success rate is still very very low. Too low. (n1 is at 30+% n3 at 50+%)

PR #209, opened 2025-02-10
Update ffxiv

silasary:
Fix for dopin putting *Eleven* progression classes in their yaml.


Eijebong:
/taskcluster merge

That's a lot better test failures wise than 0.21.0

PR #236, opened 2025-02-20
Add yoku

Eijebong:


Eijebong:
/taskcluster r+

The apworld contains itself but it's less heavy than most apworlds anyway 🤷
Reported upstream so hopefully next version will have that fixed

Eijebong:
https://discord.com/channels/731205301247803413/1225283906333446245/1342209803048456335

test failures reported here

PR #241, opened 2025-02-22
Force ooa update

Eijebong:
They keep overriding the release to make minor updates...


Eijebong:
/taskcluster merge

Same test failures as 0.3.1

PR #244, opened 2025-02-22
Add Sonic the Hedgehog 1

RePod:
Versioning subject to change based on tag naming scheme, case-sensitivity, etc.
`v0.0.2 > v0.0.3 > Sonic_1-v0.0.4 > sonic_1-v0.0.5`

It's not a good look right now in the fuzzer either.


Eijebong:
Why does the apworld contain the tracker?

Eijebong:
/taskcluster test

Might have some insight in here that would help with understanding failures

Mysteryem:
>       * Reportedly FRLG handles separate revs with separate patches

Example contents of an `.apleafgreen` patch file
image

PR #249, opened 2025-02-24
Add Final Fantasy IV Free Enterprise

thefiz:


Eijebong:
I'm a bit confused about this.

Is the apworld the one from https://discord.com/channels/731205301247803413/1170144930610557008/1341433328536129557 ?

Also I'm assuming it's going to require the upstream randomizer to be able to generate anything? Cause that's going to be an issue, last I checked that was a lot more code than I was comfortable reviewing. (plus we don't have a good story with external dependencies atm)

thefiz:
It is the APWorld from Discord, yes.  It does not require connection to an upstream randomizer during generation as far as I can tell.

thefiz:
The upstream randomizer is only called at patch time, I have verified this.

thefiz:
Closing this commit, adding a new one with your instructions.

PR #290, opened 2025-03-08
Update reventure to 0.16.2

silasary:


Eijebong:
jkx said not to update to this as generation got horribly slow IIRC.

Discord says "Version 0.14.2 is the most recent version that is stable enough to run."

Let's wait until they feel more comfortable then. Also there's direct random usage in the first file in there so it would be a no anyway

Eijebong:
There's even a 0.18.0, waiting for jkx to test it (unless you want to play that on sunday's sync and can confirm that it works?)

PR #314, opened 2025-03-17
Create frogmonster.toml

Rooby-Roo:
My version tagging is currently "v#.#.#" instead of "#.#.#" - if your code supports that for default url support let me know and I'll PR default url instead

### DIFF
--- a/index/frogmonster.toml
+++ b/index/frogmonster.toml
@@ -1,0 +1,5 @@
+name = "Frogmonster"
+home = "https://discord.com/channels/731205301247803413/1271323478779363398"
+
+[versions]
+"0.1.0" = { url = "https://github.com/Rooby-Roo/FrogmonsterAPWorld/releases/download/v0.1.0/frogmonster.apworld"}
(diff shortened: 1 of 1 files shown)


Rooby-Roo:
Note that current main version (0.1.0) is missing webworld, and I feel like if any project is gonna care about that, it's yours. I've prioritized making the rest of the world stable over the documentation for it, but expect it in a release or two.

index/frogmonster.toml
```diff
@@ -0,0 +1,5 @@
+name = "Frogmonster"
+home = "https://discord.com/channels/731205301247803413/1271323478779363398"
+
+[versions]
+"0.1.0" = { url = "https://github.com/Rooby-Roo/FrogmonsterAPWorld/releases/download/v0.1.0/frogmonster.apworld"}
```
Eijebong, line 5: You can do something like this:

```
default_url = "https://github.com/Rooby-Roo/FrogmonsterAPWorld/releases/download/v{{version}}/frogmonster.apworld"

[versions]
"0.1.0" = {}
```

The default url is just a simple replace operation on `{{version}}`

Rooby-Roo: Neat, I figured it was something like this but I needed confirmation. Commit incoming.

PR #334, opened 2025-03-30
Update ittle dew

Eijebong:


Eijebong:
/taskcluster merge

Failures reported here: https://discord.com/channels/731205301247803413/1260030530217574500/1355882417788944455

I'm pretty sure they're all false positives (empty regions not being accessible) so let's merge it. If it turns out to be an issue then I'll just revert back to 0.2.1

PR #366, opened 2025-04-04
Add Soul Blazer

Eijebong:


Eijebong:
/taskcluster test

Let's see if it would even

Eijebong:
https://discord.com/channels/731205301247803413/1209689725057761310/1358038121966473268

Reported findings upstream. Mostly concerned about the vendored library doing way more than it's used for and being way too much to review (👀 and that `eval` scares me)

PR #367, opened 2025-04-05
Add Luigi's Mansion 0.4.9

gerbiljames:


Eijebong:
Main issue here is gclib. Which sha's doesn't change in the apworld but which version's change in their releases. We don't have a way to bundle extra libs alongside apworlds atm without repacking them and we don't have automatic repacking (👀 https://github.com/Eijebong/APWM/issues/5) which would make this whole thing *painful*.

Until we have a solution for this, let's keep this open and mark it as blocked so we don't forget

SomeJakeGuy:
@Eijebong wanted to circle back on this. Was the concern for GCLib that it was required for generation or something along those lines? or that this used gclib in general? If its the former, I don't recall this ever being required for generation (or at least it was removed in one of our earliest builds), its only required for the user patching their files to create their randomized ISO.

gerbiljames:
> I think I had completely misread where gclib is used... (tbh, the file is named LMGenerator so I assumed it was on the generation side)
> 
> I'll have another look then, using it client side is "fine" by me, the lib doesn't look shady and it's just like any other mod, a risk that the player takes, not the host.
> 
> @gerbiljames Mind updating this to 0.4.4?

Donezo

Eijebong:
> Going forward though, thats the process to request an updated APWorld?

There's no "formal" process per se. If you want your apworld to be up to date in the list you can open a PR on this repo adding the new version when you release one (I keep a few older versions for each apworld so that people that used the lobby to generate can find the apworld they used for that back). You can also DM me or ping me on discord.

Eijebong:
Ah right. I'll fix the name and the home too now that's it's a channel

Eijebong:
Yikes. @SomeJakeGuy if you care to have a look at the fuzzing results. It's very very slow to generate and that failure rate is rather worrying.

Eijebong:
Funnily enough, while the output is the slowest thing in there, it's not the reason for the timeouts.

The following patch makes fuzzing roughly twice as fast which would bring the apworld into the "very slow" range instead of being "omagad what is happening" one. It doesn't fix any of the timeouts though so there must be something wrong that either ends up deadlocking, infinitely looping or gets archipelago very mad and that I cannot help with.

```diff
diff --git a/worlds/luigismansion/__init__.py b/worlds/luigismansion/__init__.py
index 49919fefa..dce225513 100644
--- a/worlds/luigismansion/__init__.py
+++ b/worlds/luigismansion/__init__.py
@@ -5,6 +5,7 @@ from dataclasses import fields
 from typing import ClassVar

 import yaml
+from yaml import CDumper as Dumper

 import Options
 import settings
@@ -889,8 +890,9 @@ class LMWorld(World):

         # Output the plando details to file
         file_path = os.path.join(output_directory, f"{self.multiworld.get_out_file_name_base(self.player)}.aplm")
+
         with open(file_path, "w") as f:
-            f.write(yaml.dump(output_data, sort_keys=False))
+            f.write(yaml.dump(output_data, sort_keys=False, Dumper=Dumper))

     # TODO: UPDATE FOR LM tracker
     def fill_slot_data(self):
```

BootsinSoots:
The only thing I could think of that causes generate_output to run slowly is the threading events I have set up to deal with by-sphere items like boo_health and hints  - but that's to reduce the time already. What's the criteria for these timeouts? I've never had generation time out on me, so I assume it isn't just an innate timeout.

Eijebong:
It's a single generation with one random YAML taking more than 10s.

I think the output being slow is a red herring because making generate_output ~10x faster (with the patch I attached) didn't resolve the timeouts. So either those generations are deadlocked or are looping. I can check later if I can reproduce them locally but you should be able to do that yourself by getting the archive in the CI job, taking one of the timeout cases, placing the associated YAML in your Players folder and using `Generate.py --seed <the seed that you can find in the associated log>`

Eijebong:
This is interesting because I cannot reproduce locally outside of the fuzzer 🤔 Makes me wonder if you're leaking things in a global state somewhere that would break generation (and be an issue if you have multiple LM in the same multiworld). Because it also happens with only one worker process and seems to happen at regular intervals.

`I..FF..F..I..FT.F..I.II...T.IFI.FF.IFII..I.FI.I.T....FI..FI.FI.T...I.FIIIIIIIF.......TI.....F...ITII...II.III.I.II.FF..TIII.I.F.II.T..F..I..F.III...IF...FIFT.FII.III.........I.T...FF...I.FI...II.FI.FTFI..II...I...I.FII.IT` every `T` is a timeout.

Positions are `[14, 26, 48, 63, 85, 97, 119, 131, 156, 176, 199, 220]`

Which has a lot of `12` diffs in it...

Eijebong:
Narrowed it down to `get_hints_by_option`

SomeJakeGuy:
@gerbiljames just released 0.4.5 with a bunch of changes from Boots / Eije the other day if you can bump that version and gives these tests another go.

Eijebong:
> if you can bump that version and gives these tests another go.

@SomeJakeGuy https://github.com/Eijebong/ap-actions/ *wink* *wink* You could have those too, on every push, on every PR. Just sayin'. Full example here https://github.com/gerbiljames/Archipelago-Crystal/tree/future/4.0.0/.github/workflows

SomeJakeGuy:
@gerbiljames @Eijebong I appreciate you both, thanks so much for this. I will work on getting those github actions created because that will definitely be great to test again 

(apologies on the delay, work kicked my ass last few days and I was so tired to go from coding as a day job to more code lol)

Eijebong:
Don't feel pressured or anything, there's no rush whatsoever on my side. I've had games in PRs here for 6 months 🤣 it'll happen when it happens. Also don't hesitate to ping me on discord if you need help with CI

SomeJakeGuy:
Nah i have too much of a passion for this, genuinely. Ever since I picked this to randomize its been a blast, I just dont want you to feel ignored <3

SomeJakeGuy:
That and I want to play this in the next Chrism AP, so I gotta get this all fixed hahahaha

Eijebong:
Not sure why the diff was hallucinating changes but I rebased because I did change some CI related stuff recently

SomeJakeGuy:
I finally got out of my work week hell I endured, so I'm working on getting GitHub actions setup so I can re-test these on our end before we provide the next version.

SomeJakeGuy:
Also my last triple run output is "1000 / 1000 done. 8 failures, 0 timeouts, 579 ignored.", why is the ignored so high and what does it even mean in this context? I assume thats option errors?

Eijebong:
> Also my last triple run output is "1000 / 1000 done. 8 failures, 0 timeouts, 579 ignored.", why is the ignored so high and what does it even mean in this context? I assume thats option errors?



> Also my last triple run output is "1000 / 1000 done. 8 failures, 0 timeouts, 579 ignored.", why is the ignored so high and what does it even mean in this context? I assume thats option errors?

Yeah ignored are OptionErorrs. I don't have a set threshold because it really depends on the errors found. I'd be comfortable enough including this with 8/421 failures on 3 YAMLs.

SomeJakeGuy:
Okay I'm comfortable with these outputs now. Go ahead and pull down 0.4.7 and we should be in a much better place

gerbiljames:
Will update today when I get a chance (or @Eijebong can do it)

SomeJakeGuy:
Okay we did finally discover the issue. It was the classic issue of CRLF vs LF. I will make a new patch for that.

SomeJakeGuy:
I assume it won't cause an issue if I up GCLib's version (which isn't used in generation still), but I did see they have a new version that "specifically fixes windows compilation".

SomeJakeGuy:
Also V0.4.8 is now released https://github.com/BootsinSoots/Archipelago/releases/tag/V0.4.8

SomeJakeGuy:
Damn only 54 failures out of 10000, nice

SomeJakeGuy:
Okay, https://github.com/BootsinSoots/Archipelago/releases/tag/V0.4.9 confirmed to have fixed the new-line issues in one of the settings for both Linux and Windows. And now we are linux compatible :D

Eijebong:
/taskcluster r+

🥳 Thank you for dealing with all this and getting it into a mergeable state

SomeJakeGuy:
Also @Eijebong I had the wrong AP world attached it seems, so I edited the release of 0.4.9 as a heads up

PR #396, opened 2025-04-16
Re-add Yacht Dice 2.2.2 beta

Mysteryem:
The 2.2.2 beta was removed in the update to AP 0.6.1, but the 2.2.2 beta is ahead of the AP 0.6.1 release of Yacht Dice.

If it was removed for other reasons, I was not aware of them.


Eijebong:
/taskcluster r+

It was removed because I had wrongly assumed it got merged 🙃
Should've guessed

PR #401, opened 2025-04-17
Update Ape Escape to last version

Thedragon005:
This version was tested for a while and had at most 2/1000 errors in fuzz.py error when genereating and no user reported errors


Eijebong:
/taskcluster r+

Thanks :)

PR #403, opened 2025-04-17
Add kirby super star

Eijebong:


Eijebong:
/taskcluster fuzz -r10000 -n3

restrictive starts?

Eijebong:
/taskcluster r+

Error reported here: https://discord.com/channels/731205301247803413/1170152771043999764/1362528226471051446
Merging as the only failures fail very early in `generate_early`, in dumb YAMLs so the checker will catch those

PR #407, opened 2025-04-18
Create ttyd.toml

earthor1:


Eijebong:
Need to figure out gclib first (which would also unblock Luigi's mansion). Probably will end up with the current latest version of it and hope it works with both worlds...

earthor1:
Ah understood, I remember this fuzzed quite well so assumed it was ready to rock and roll. I'm curious why gclib can't be just packaged with the apworld file itself and requires a user to go through the steps.... Interesting.

benny-dreamly:
> Ah understood, I remember this fuzzed quite well so assumed it was ready to rock and roll. I'm curious why gclib can't be just packaged with the apworld file itself and requires a user to go through the steps.... Interesting.

yeah that's what the core people keep saying is the "right" way to do it, but uh... apparently nobody does that...

earthor1:
From speaking with the apworld dev, it's apparently not possible to package it within the APWorld. Maybe one day gclib can be added to core so that it becomes a moot point.

Eijebong:
I think just like LM, this isn't blocked by gclib anymore

Eijebong:
All failures are the same, reported here https://discord.com/channels/731205301247803413/1205136944259145728/1378709694088089772

PR #422, opened 2025-04-21
Update fe8

Eijebong:


Eijebong:
This is wrong, it should be properly detected as `0.3.3-rc1`

PR #425, opened 2025-04-21
Update openrct2

Eijebong:


Eijebong:
This should be detected as `0.1.17-beta`

PR #430, opened 2025-04-21
Update outer_wilds

Eijebong:


Eijebong:
This should be detected as 0.3.7-1

PR #434, opened 2025-04-21
Update checksmate

Eijebong:


Eijebong:
This is very wrong, but not a bug per se. They have some very old versions using the date as the version, of course 2023.x.z is way greater than 0.2.x...

PR #444, opened 2025-04-24
Add Lunacid

Eijebong:


Eijebong:
/taskcluster r+

11/10000 for n1, I'll take it. Reported failures here anyway (those `KeyError` should probably not happen): https://discord.com/channels/731205301247803413/1176614031067455528/1365014623740235797

PR #451, opened 2025-04-27
Add Celeste (Open World)

pyrox0:
This is a new Archipelago mod for Celeste that makes it more in the vein of a traditional randomizer, having interactions with some interactable objects locked behind checks and locations. It's highly customizable and very fun to play!


Eijebong:
/taskcluster r+

It's trying to compete for slowest apworld in the index but other than that look ok.

PR #452, opened 2025-04-28
Update wordipelago

darkfire006:


Eijebong:
/taskcluster fuzz -r10000 -n1

curious what that generate_early removal will do to the success rate

Eijebong:
/taskcluster r+

🤷 It probably was detecting restrictive starts a bit too much. Fuzz on -n3 is at 2/10000 failures which is fine.

PR #460, opened 2025-05-04
Add simpsonshitnrun

Eijebong:


Eijebong:
/taskcluster merge

Don't care about test failure, fuzzing is fine even if slow but that's to be expected as this is basically a manual

PR #463, opened 2025-05-05
Add Digimon World

Eijebong:


Eijebong:
(no r+, it's using random directly, reported directly to Arson)

PR #464, opened 2025-05-07
Update zork_grand_inquisitor.toml

Fore772:
Added V2 Beta, prepare versioning for when removed from core-verified.


Fore772:
Wasn't sure what to do about the Supported tag as of now.

index/zork_grand_inquisitor.toml
```diff
@@ -1,3 +1,8 @@
 name = "Zork Grand Inquisitor"
-home = "https://archipelago.gg"
+home = "https://discord.com/channels/731205301247803413/1224389493289648218"
 supported = true
+default_url =  "https://github.com/SerpentAI/Archipelago/releases/download/zgi-v{{version}}/zork_grand_inquisitor.apworld"
+
+[versions]
+"1.1.0" = {}
```
Eijebong, line 7: Why do we need 1.1.0? I keep old versions on the lobby but only if they were present before, I don't intend on making an apworld museum

Fore772: the 1.1.0 is the version currently in 0.6.1, I think. the newest version is not backwards compatible at all.

Eijebong: According to the diff 1.1.0 is different than the version in 0.6.1.

Fore772: Weird. It behaves I think the same as how the supported one works, it may have had changes in the official branch that had it work correctly. Does the "supported = true" section pull in the core-verified version then? In which case just setting the core-verified version as default but still adding the beta do what I was intending?

Eijebong: Yeah, having `supported = true` pulls in the core verified version specified here https://github.com/Eijebong/Archipelago-index/blob/main/index.toml#L2 in the index.

Fore772:
I know for a fact I did that whole process wrong.

Eijebong:
Welp, no idea why it's failing x) Try `"0.6.1"` instead I guess, we'll see

Fore772:
> [Review changes](https://apdiff.bananium.fr/EHln247dSbWUqk1OUYzEKg)

Did this one technically work? Not familiar with reading the diff, but I feel like it didn't pull in the 0.6.1 version.

PR #480, opened 2025-05-10
Create rusted_moss.toml

Fore772:
Add rusted_moss.toml.


Eijebong:
/taskcluster fuzz -r 10000 -n1

I doubt it but let's see

Eijebong:
/taskcluster r+

Still slow but at least workable. It'll probably need an update once they fix their release to be an apworld and not a zip file. Reported here https://discord.com/channels/731205301247803413/1239286100472758302/1373254424121839716

PR #483, opened 2025-05-10
Add manual for Lego Star Wars: The Complete Saga

Mysteryem:
This is a repack of Roush's v7.1 Lego Star Wars: The Complete Saga manual world:
https://discord.com/channels/1097532591650910289/1156373916370083880/1290558889963552789

The repack uses my work-in-progress fork of manual that improves logic performance. Because this fork is based on neither a stable nor unstable release of manual, the included manual client is given the lowest possible priority, causing manual clients from other loaded manual worlds to be loaded in preference to the client included in this world.

---

I think it could be worth giving the version an extra identifier rather than just `0.7.1` to help indicate that this uses a potentially unstable version of manual (more so than the official 'unstable' releases), I've seen `-alpha+unreleased` used for Celeste for example.

### DIFF
--- a/index/manual_legostarwarsthecompletesaga_mysteryem.toml
+++ b/index/manual_legostarwarsthecompletesaga_mysteryem.toml
@@ -1,0 +1,8 @@
+name = "Manual_LegoStarWarsTheCompleteSaga_mysteryem"
+display_name = "Manual: Lego Star Wars The Complete Saga"
+home = "https://discord.com/channels/1097532591650910289/1156373916370083880"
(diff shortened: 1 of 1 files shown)


Eijebong:
Test failure I don't care about. I do worry a bit at both the failure and timeout rates though

Eijebong:
/taskcluster fuzz -r1000 -n3

Urgh I thought I had implemented timeouts on this... Guess not. Let's hope we still get some results

Eijebong:
/taskcluster r+

Holy moly that's fast now

PR #484, opened 2025-05-11
Update Ape Escape

Thedragon005:
For 0.8.4 update


Eijebong:
I usually keep a few versions for each apworld in case someone has generated their room with it, to not invalidate their list of apworlds. Can you keep the old version and add the new one instead of changing it?

Thedragon005:
Sure thing, remade a commit with the 0.8.2 version.
Thanks for the info :)

PR #489, opened 2025-05-14
Update reventure

Eijebong:


Droppel:
Hey I've been alerted to this issue by jkxwing on discord.

As far as I can tell the fuzz test generates yamls with random settings. I assume this sometimes turns on the "experimentalRegionGraph" setting. That setting uses a highly experimental way to generate the logic for the game (In theory allowing it to do some fun things). But in practice this currently leads to very long generation times (Easily 30 minutes on a decently powerful CPU) and is still likely to generate broken logic.

The culprit here is probably the long generation time leading to the timeouts. If there is a way to disable this setting for fuzzing that should solve the problem. (And of course disallow using it in any way in any AP Game that is not specifically meant to test this setting).

PR #494, opened 2025-05-17
Add RE3R

alex73630:


Eijebong:
The apworld doesn't even generate its own template, reported that, the 10k failures and the use of bare Exceptions here: https://discord.com/channels/1085716850370957462/1261562909809709056/1373625513444380743

PR #495, opened 2025-05-17
Update RE2R to 0.2.6

alex73630:


Eijebong:
I usually keep a few older versions so that people using those on the lobby can easily get them back. Could you keep 0.2.5 in the list of versions? Should be as easy as

```toml
[versions]
"0.2.5" = {}
"0.2.6" = {}
```

alex73630:
Sure! Just added the version back ;)

Eijebong:
/taskcluster fuzz -r 10000 -n 1

Unfortunately that's probably going to get very angry because of that `Exception` raised, we'll see how that goes

Eijebong:
/taskcluster r+

> 10000 / 10000 done. 9141 failures, 0 timeouts, 0 ignored.

Yeaaaah... AFAICT, all failures except 4 are those two exceptions though so it's "fine".
I reported both the fact that they should be using `OptionError` and the gen failures here: https://discord.com/channels/1085716850370957462/1259648589039272028/1373622654812033136

PR #496, opened 2025-05-18
Update balatro.toml

dopinpt:
no vroom vroom so pls gib funny poker chrisOoga


Eijebong:
/taskcluster merge 

Nothing new

PR #497, opened 2025-05-18
Update Sonic Heroes to 1.3.0

TrashMouthColt:
🙏 it should be able to be tested now


Eijebong:
Reported the fuzzing regression here https://discord.com/channels/731205301247803413/1234592520403423352/1373969340923117648 we'll see.

PR #501, opened 2025-05-21
Create bingo.toml

Fore772:


Fore772:
Bingo is a weird meta game/side game, in that it would be impossible for it to goal on its own. It requires a second game for it to work.

Eijebong:
Yeah... I can see that from the tests. I really wish people wouldn't 🙃 If you don't want to play AP, don't play AP.

There's a bug in the fuzzer and in CI though, it shouldn't fail like this. And when it fails the task should be marked as such

PR #507, opened 2025-05-22
Update pmd_eos.toml to 0.3.0rc5

Fore772:


Eijebong:
/taskcluster fuzz -r 1000 -n 1-3

I think tests deadlocked?

Eijebong:
The test does deadlock yeah :) Reported here: https://discord.com/channels/731205301247803413/1175911373822242946/1375223202065678406

PR #509, opened 2025-05-22
Update reventure.toml to 0.18.2

Fore772:


Fore772:
Supercedes PR #489

Eijebong:
/taskcluster fuzz -r 100 -n1

Probably gonna be the same result as #489

Eijebong:
Yeah I wrote it wrong. I meant that it couldn't be enabled in a YAML at all by default (either by having a host.yaml option or by just deleting the option completely). I run those apworlds while validating YAMLs or generating rooms. You can probably understand why I wouldn't want a single YAML to have one option and have a validation take 30mn 🙃

Droppel:
I absolutely understand! So if there were a host.yaml option to prevent the setting that would work for you? Because if yes, I can look into that. Alternatively I could also publish two different APWorlds, one with the setting and one without.

Eijebong:
Yeah, either option would be fine for me.

Droppel:
Just released https://github.com/Droppel/ReventureEndingRando/releases/tag/0.18.1 which adds a host.yaml setting that needs to be true to allow using the experimental setting. It is disabled by default. (Not sure if you want this in a new PR I'll leave that up to you, but feel free to tag me if any further issues do arise)

Eijebong:
@Droppel ^ that test failure is a blocker

PR #512, opened 2025-05-22
Update sonicriders.toml to 0.2.2

Fore772:


Fore772 (commented):
Right, it doesn't exist yet. Should I be copying the apworlds in myself?

Eijebong:
Yes, it's not going to appear magically :p

PR #527, opened 2025-05-26
Create waterywords.toml

Fore772:


Eijebong:
Reported directly to Spineraks. This is too slow.

PR #528, opened 2025-05-26
Create rabi-ribi.toml

Fore772:
Trying a [version] that I don't think will be accepted first.
edit: It didn't work, updated to correct versioning.


Fore772:
What I get for doing these on the work computers with Japanese keyboards.

Eijebong:
/taskcluster fuzz -r 500 -n 3

Maybe it's just restrictive starts given the error...

PR #539, opened 2025-05-30
Update zork_grand_inquisitor.toml to 2.0.5

Fore772:


Eijebong:
/taskcluster r+

Nothing alarming in the fuzzing results

PR #574, opened 2025-06-02
Create bingo.toml

Fore772:


Eijebong:
#501 for reference. I'm still not sold

Fore772:
Was the bug in the fuzzer and CI reported to the dev, or are you against the apworld as a concept?

Eijebong:
I didn't report it because there's nothing to report AFAIU, this is juts not a "game" that can generate on its own so it will always fail tests, always get all failed in the fuzzer. And yes I'm also not sold on the concept, the more it goes the more I hate these meta games.

If you don't want to play archipelago then go out and play some bingo instead, I'm sure your nearest retirement home would love having young people engage with people there.

PR #578, opened 2025-06-02
Create ufo50.toml

Fore772:


Mysteryem:
In the Barbuta game, it looks like there is a missing indirect condition for the `"Wand Trade Room"` region and the `"Starting Area" -> "Mimic Room"` entrance, though it seems to be the only one in the apworld and the rest of the apworld looks OK at a glance.

```python
    def has_wand(state: CollectionState) -> bool:
        return state.has(wand, player) and state.can_reach(regions["Wand Trade Room"])
```

```python
    regions["Starting Area"].connect(regions["Mimic Room"],
                                     rule=lambda state: has_wand(state))
```

PR #579, opened 2025-06-02
Create waterywords.toml

Fore772:


Eijebong:
#527 for reference. This was waaaaay too slow to gen

PR #585, opened 2025-06-03
Add Oxygen Not Included

Fore772:


Eijebong:
Looks like the version given is the wrong one, from the old repo. New releases are at https://github.com/ShadowKitty42/ONI-Archipelago/releases/

Eijebong:
And the fuzzing results are worse with n3. Had a quick look and it looks like it's hitting `AssertionError:  - 1 already exists in the location cache.` a lot. So probably leaking stuff in global state.

Eijebong:
Reported here https://discord.com/channels/731205301247803413/1209238808625946634/1380145413423890544

PR #587, opened 2025-06-03
Add Grim Dawn

Fore772:


Eijebong:
/taskcluster merge

Reported here: https://discord.com/channels/731205301247803413/1114646643225137224/1379895426664108194

All failures in fuzzing are because of an Exception that should be an OptionError, otherwise it's all good. That's not a blocker

PR #598, opened 2025-06-10
Add Mindustry

Fore772:
Related to PR #531


Eijebong:
A bit slow, will have to check if there's something obvious

PR #603, opened 2025-06-10
Re-add Super Junkoid

Fore772:
Because of the AP rules, I know this one now treads a line. Here is a blurb from the repo as well:

"Due to this apworld containing derivative work if you want to share this apworld you must not post the file directly in the main Archipelago discord server and instead post a link to this github release"


Eijebong:
I see...

PR #620, opened 2025-06-14
Add powerwash simulator

Eijebong:


Eijebong:
Reported here https://discord.com/channels/731205301247803413/1158367507703410739/1383396652953501727

Eijebong:
/taskcluster r+

Errors are mostly if you don't enable any locations, should be an OptionError but it happens in generate_early which is good enough for me. A few FillErrors but nothing *too* concerning

PR #624, opened 2025-06-15
Update tcg_card_shop_simulator.toml

Muuffin071:


index/tcg_card_shop_simulator.toml
```diff
@@ -3,5 +3,5 @@ home = "https://discord.com/channels/731205301247803413/1302063649812250665"
 default_url = "https://github.com/FyreDay/Archipelago-TCGCardShopSimulator/releases/download/v{{version}}/tcg_card_shop_simulator.apworld"
 
 [versions]
-"0.1.1" = {}
```
Eijebong, line 6: Keep the old version please, there might be people running asyncs generated on the lobby using this (although I'd hope not given the version difference) and I would hate for their room to start distributing the wrong version

Muuffin071: gotcha and will do.

Eijebong: Thanks!

PR #637, opened 2025-06-19
Update th06

Eijebong:


Eijebong:
Reported here https://discord.com/channels/1085716850370957462/1353408740480778270/1385241878106210324

Fuzzing rate took a nosedive from 5 in 10k with 3 YAMLs per gen to 2741 in 10k

PR #647, opened 2025-06-20
Update STS

Eijebong:


Eijebong:
Reported here https://discord.com/channels/731205301247803413/884926940182609920/1385567973141647371

Eijebong:
/taskcluster r+

Ok, even with n3, not a single gen failure that isn't early, even if the exceptions are wrong, this doesn't matter for the lobby itself, the user facing error will just be a bit weird

PR #449, opened 2025-04-26
Update candybox2

Eijebong:

PR #690, opened 2025-07-04
Update candybox2.toml

palex00:


Eijebong:
Dupe of #449 

Failure rate is way too high `[task 2025-06-02T10:55:59.068+00:00] 10000 / 10000 done. 631 failures, 1 timeouts, 0 ignored.`

PR #698, opened 2025-07-08
Update autopelago.toml to 0.10.4

Fore772:
Seems to mostly be performance fixes.
Previous PR opened from wrong branch, sorry.


Fore772:
And I did the wrong order... Reopening as this is the correct PR.

PR #703, opened 2025-07-08
Update tyrian

Eijebong:


Eijebong:
> Fill.FillError: Not enough filler items for excluded locations. There are 1 more excluded locations than excludable items.

This feels like actual bugs, a world shouldn't be submitting more items than locations to the pool. It's only passing with the second apworld because it's removing fillers from that world

Mysteryem:
Personally, I think these errors are something a world could consider checking in advance if it has chosen to implement options for excluding locations and/or increasing the percentage of the item pool that is non-excludable, and can run into the issue of more excluded locations than excludable items.

If every apworld were to use the excuse that chosen options are not a problem when there are other slots in a multiworld, then all it takes is too many slots relying on other worlds to bail them out, and then generation fails.

---

On a separate note, while I'm here, I will mention that Tyrian does have some very rare logic issues where collecting an item causes locations to go from in-logic to out-of-logic. It is possible that these logic issues could be the cause of some of the general FillErrors that occur when fuzzing Tyrian on its own.

I ran 20000 solo generations with the fuzzer, and out of 15373 successful generations, 18 failed the [fuzzer hook](https://github.com/Mysteryem/Archipelago-fuzzer/blob/mysteryem_hooks/hooks/collect_accessibility_test.py) I wrote to check for accessibility reducing after collecting an item. The failures are attached (the "swapping" prints you may see in the logs are uncommited changes that I forgot to remove, that just print "swapping" when fill has to swap)
[tyrian.zip](https://github.com/user-attachments/files/25136806/tyrian.zip)

The error messages from the hook failures show what item being collected caused which locations to go from in-logic to out-of-logic, show the value of `state.prog_items[1]` before and after that item was collected, and show all previously collected items and the order they were collected in.

Eijebong:
/taskcluster r+

I don't want to block this forever on that but I really wish you'd consider throwing option errors for those options that produce too many progression items... If everyone was just "nah it's fine other worlds will fix the issue" we'd never get anywhere

PR #727, opened 2025-07-17
Update jakanddaxter.toml (new home)

massimilianodelliubaldini:
Don't know what to do for the versions now but LMK and I'll put in another commit.


Eijebong:
Done as part of the 0.6.2 update, thx

PR #732, opened 2025-07-17
Replace manual Lego Star Wars: The Complete Saga with the full apworld

Mysteryem:


Mysteryem:
I found some logic bugs with bonus levels, so I'll release a new version with that fixed.

PR #747, opened 2025-07-28
Updated Spyro 3 to 0.5.2

Cevvin7:


Eijebong:
A lot of these happen post gen :/ There's only 2/10k errors that I'd consider a problem, but those hints one happen way too frequently and post gen

@ArsonAssassin I know you're taking a break right now, just pinging you for when you find time/motivation

Eijebong:
Yeah, that is the intended behavior. And if you were leaking data between worlds it could also probably happen when you have multiple spyro 3 in the same multiworld. The screenshot that you sent looks like you got a perfect score now :) Thank you for looking at it.

Eijebong:
/taskcluster r+

Thanks a lot for taking care of those fixes

PR #764, opened 2025-08-05
Add cuphead

Eijebong:


Eijebong:
/taskcluster r+

Probably gonna regret this. All fuzzing errors apart from the few fill errors are caught early on, early enough that the YAML checker will complain about them. (they really should be `OptionError` and the "0 isn't a valid mode" stemming from the apworld massaging options wrong isn't exactly inspiring confidence but we'll see...

PR #778, opened 2025-08-09
Added Civilization V to the index

1313e:
This adds the 0.6.0 release of the Civilization V APWorld to the index.

### DIFF
--- a/index/civv.toml
+++ b/index/civv.toml
@@ -1,0 +1,6 @@
+name = "Civilization V"
+home = "https://discord.com/channels/731205301247803413/1342924294757552229"
+default_url = "https://github.com/1313e/Civ-V-AP-World/releases/download/v{{version}}/civv.apworld"
+
+[versions]
+"0.6.0" = {}


1313e:
- Fixed the empty range error (I did not get the unique trap names when comparing if someone blacklisted all trap items).
- Looking into the unbeatable issue. Cannot seem to reproduce that locally when using the YAMLs from the fuzzer. 🤔

1313e:
Uhm, I think there is a generation problem somewhere.
I took the YAML from run 7040 (as the fuzzer output stated a location was unreachable even though it also reported that the sole item required for that location was on the very first check in the game and thus that is BS) and generated it locally with the same APWorld and AP version, and it generated just fine.
Seed was set to be the same as used for that run in the fuzzer as well.
Output is below:
[Generate_727626256_2025_08_10_14_29_48.txt](https://github.com/user-attachments/files/21705301/Generate_727626256_2025_08_10_14_29_48.txt)

Do you have any ideas what could be the issue here?

1313e:
@gerbiljames The two files you sent have the exact same randomization.
Each location has the same item in both generations.
Therefore, the seeds are the same.

Their playthroughs being different is because there are multiple paths to victory.


However, this still does not explain HOW so many generations are failing on the fuzzer.
I have yet to see any generation fail. Ever.
At this point, I have probably generated a 1000 seeds during development, and not once did it fail because it was unbeatable.
Yet, in the fuzzer, it fails 50% of the time.

Additionally, both myself and James have now confirmed that the generation succeeds using the 7040 YAML and seed.
Why is it failing in the fuzzer then?

gerbiljames:
It does seem to fail more often the later into the run it is, are you keeping some global state somewhere?

gerbiljames:
It seems to be both, there's some state that's leaking globally

I can get a consistent failure with the following on your `develop` branch

[civv_n20.zip](https://github.com/user-attachments/files/21741217/civv_n20.zip)

1313e:
@gerbiljames Okay, that one also gives me a failure.
I finally have a scenario to debug \o/
Thanks.

1313e:
@gerbiljames Okay, I finally found what the issue was. 🤦 
Thanks for the help, man. :)

PR #786, opened 2025-08-12
Create pokemon_bw

mzweiten:
may not work, but they tried really hard to get the kinks worked out from 0.1.0


mzweiten:
I'm going to keep this PR open just in case they fix things soon(TM). Don't want to duplicate work later.

gerbiljames:
@mzweiten 0.2.1 looks good for fuzzing, but did fail one test for me

PR #792, opened 2025-08-13
Update CrossCode

kingdragonite55:


Eijebong:
/taskcluster fuzz

Definitely not losing my mind 😭

index/crosscode.toml
Eijebong:
```suggestion
```

index/crosscode.toml
```diff
@@ -10,3 +10,4 @@ disabled = true # 8% failure rate, update is the same
 "0.6.6" = {}
 "0.6.8" = {}
 "0.6.11" = {}
```
Eijebong:
```suggestion
```

CodeTriangle:
Hello,

I am the maintainer of the CrossCode mod and APWorld. I was informed of test failures with the mod. I have since resolved these.

I do not know about the scope or structure of your project here, but I also can't help but notice what seems to be a kind of replacement of the 0.7.x line of the APWorld with the 0.8.0-pre.X line. While power users may desire the ability to generate on the prereleases, they are generally not considered stable for generation in large multiworlds. Since you are offering downloads to the apworlds, I think it is relevant to consider this.

CodeTriangle:
Well. It does seem that I broke determinism. I assumed the problem was just my failing unit tests.

Give me a bit to fix that.

Eijebong:
determinism isn't the main issue tbh, those are just useful for us to know if we can trust UT or not. 5% of fill errors is a much more concerning problem to me. (even if options in those YAMLs might not make sense in which case it should be raising `OptionError` in `generate_early` instead of letting fill fail after 2h because someone either sent a malicious YAML with the intent of making gens fails or made a mistake)

CodeTriangle:
I take it back. It seems that there is one specific options combination that causes the most problems which I should probably block anyway.

Eijebong:
> In practice the incidence of fill errors is nowhere near 5%

Yeah for sure, with settings that do make sense the error rate surely isn't that high. My issue is that when organizing open syncs where more than 1000 people join, including a lot of people for whom it's their first AP, we cannot review every single setting of every single YAML. And if a world ends up failing in fill after 2h because of those settings instead of saying "your options make no sense, fix them", then we have a very bad time. Hence why I'm "strict" on random failures like this for inclusion.

CodeTriangle:
> I think you should raise some option error when `consumable_weight` + `drop_weight` are 0 and when all `rarity_weights` are 0? Since both these end up with `_filler_pool_weights` being all zeros.

Good call, I hadn't considered what would happen with that. Those options are not used very often so I'm not surprised those slipped through the cracks.

> Way more concerning to me, we'll see how much the `So'najiz Key` hates me in the future I guess, but I didn't like it too much when I set that as a meta yaml for fuzzing with 10 YAMLs per gen (goes up to ~5-6% failure rate):
> 
> ```yaml
> CrossCode:
>   element_shuffle: own_dungeons
>   small_key_shuffle: original_dungeons
>   chest_key_shuffle: own_world
> ```

These three have always been the most unstable options in the randomizer. Forcing lots of items into the dungeons leads to a fair amount of instability. I've considered removing them but I think they serve an important purpose. The dungeons are restrictive environments to fill, but it's usually offset by the openness of the overworld. There might be some world-specific trickery I can pull off here. I can ask the Archipelago world dev chat about methods to tackle it.

> If that ends up being a pain I might disable this again until it gets fixed, I would really like to avoid another alttp situation where a single world ruins half of my prefills (although it is played less and is less likely to fail than alttp was)

It looks like you merged it in already, but if it causes you trouble again, feel free to mention me here, or DM me on Discord (my handle is the same as my GH username). This conversation has actually been extremely helpful, and it's given me a valuable tool to use in my testing, so I'll be more than willing to address your concerns as hearing you out has so far only made my world better.

However, if you want my honest take, I truly doubt you have anything to worry about. Like I said earlier, these options are rarely touched, and for the three years that the `*_shuffle` options have been available, I can count on one hand the number of Fill Error reports I've gotten related to them. There's always the trusty `meta.yaml` trick if all else fails, plus I'll be using the fuzzing tool to check my work from now on.

PR #797, opened 2025-08-13
Add rogue_legacy 1.1.0

gerbiljames:


Eijebong:
/taskcluster r+

All failures happen in generate_early and are Exceptions raised instead of OptionError so i'll take it

PR #803, opened 2025-08-14
Update pmd_eos.toml to 0.3.1

OceanWave101:
(Side note, they provided a separate client.py file with this release to (manually) fix some special characters; not sure how you'd want to handle that, if at all)


Eijebong:
/taskcluster r+

I couldn't find that modified client on discord but from what I understood of the issue, as long as your slot name doesn't contain a `~` then it should be fine

PR #818, opened 2025-08-16
Update factorio_saws.toml

mzweiten:
another AP request while community sync prep was underway. I hope commenting out works on the 0.0.6 version.


index/factorio_saws.toml
```diff
@@ -1,7 +1,6 @@
 name = "Factorio - Space Age Without Space"
 home = "https://discord.com/channels/731205301247803413/1311946326040383589"
-disabled = true # https://discord.com/channels/731205301247803413/1311946326040383589/1359608891943551156
+default_url = "https://github.com/silasary/Archipelago/releases/download/factorio_saws-{{version}}/factorio_saws.apworld"
 
 [versions]
-# https://discord.com/channels/731205301247803413/1311946326040383589/1329822970096521399
-"0.0.6" = { local = "../apworlds/factorio_saws-0.0.6.apworld" }
```
Eijebong, line 7: Remove that file from the repo too

Eijebong:
Oh I see tests are testing for GERpocalypse on their own, don't need anything else then 🤣

@silasary ^ This is a blocking failure. Basically will fail to gen if anything in the MW uses GER. You should be able to repro locally by generating it with kh1 with a default YAML

PR #826, opened 2025-08-22
Update Simpsons Hit and Run to recent beta version

mzweiten:


benny-dreamly:
pretty sure the 0.2.4 beta should be good

PR #833, opened 2025-08-25
Fix URL for 0.1.16-beta in openrct2

BroOtti:


Eijebong:
That's what I get for not pulling before testing :) I see it now #825 broke this.

Eijebong:
Thx

PR #839, opened 2025-08-28
Create pokemon_platinum.toml

mzweiten:
no need to run fuzzing tests yet


Eijebong:
Is this good enough yet?

PR #844, opened 2025-08-30
feat: Added v0.6.2 and v0.6.3 releases to Civ V

1313e:


Eijebong:
/taskcluster test-fuzz

Not sure why 0.6.2 is in there, I usually don't take intermediate versions, but now I'm curious if CI behaves when this happens

1313e:
I put it in as I had forgotten to add 0.6.2 to the index.
It felt a bit weird to have 0.6.1 and 0.6.3 in there, but not the version in between.

Nice to see that that works fine as well. :)

PR #850, opened 2025-09-04
Updating TTYD to 0.5.1

earthor1:


earthor1:
Dev notified, this is fixed in 0.5.1. I'll update PR once that update is released so it can be retested.

earthor1:
Good news: apworld is updated and was fuzzed successfully by the dev (but should obviously get fuzzed again here)
Bad news: it now has deathlink -_-

PR #851, opened 2025-09-05
Update spyro2.toml

Uroogla:
Bump Spyro 2 to 0.3.2.


Uroogla:
In my own fuzz testing, I got fill errors about 0.03% of the time for valid settings over 100,000 seeds.  This dropped to 0 over 10,000 when 2 YAMLs were included, so I believe they're restrictive starts.  Continuing to look into the different classes of these, but wanted to flag this in advance.

Eijebong:
/taskcluster r+

thx

PR #859, opened 2025-09-09
Add Haste

RePod:
~~`default_url` does not point to an actual file. The GH Release is a zip containing the apworld, YAML, license, and README.~~

Fuzzing and general apworld usage *may* fail with `Menu already exists in region cache.`
Encountered with `Archipelago (0.6.3), Python 3.12.10` (source)
Not encountered witth `Archipelago (0.6.3), Python 3.12.10 (frozen)` (a typical install)


RePod:
Bad news for the 3 Haste players.

MoSadie:
well that's an unexpected result, looks like the test fuzz died in 13 seconds and threw multiple exceptions... I didn't see that coming lol

PR #862, opened 2025-09-10
Add Pikmin 2

kingdragonite55:
chrisWOW


kingdragonite55:
@Eijebong This version should fix the issues with the previous apworld. Let me know if anything else is awry.

palex00:
0.7.0 is now released. Might be worth replacing 0.6.1 with that and seeing how it fares @kingdragonite55

Mysteryem:
0.7.1 should have fixed the previously mentioned issues

PR #896, opened 2025-09-13
Update tevi

Eijebong:


kingdragonite55:
Tevi's apworld's version number is separate from the version number of the randomizer mod, 0.5.5 is the latest version of the apworld (bundled with randomizer ver. 1.3.0). Tevi is already up to date.

Eijebong:
I see, thanks. I use a script to open those (quite obviously given the spam), I'll just exclude tevi from that from now on

PR #935, opened 2025-09-19
Added Pokemon BW ap-world in index

MrMOrak:


index/pokemon_bw.toml
```diff
@@ -0,0 +1,6 @@
+name = "Pokemon Black and White"
+home = "https://discord.com/channels/731205301247803413/1414928022854696991"
+default_url = "https://github.com/BlastSlimey/PokemonBWAP/releases/download/{{version}}/pokemon_bw.apworld"
```
Eijebong: Pain. I'm not adding a 32MB apworld to the index. And the one without map is an apworld in a zip which... Also no. And I would really like to avoid having the apworld in tree

PR #936, opened 2025-09-21
Update manual_residentevil4_vincentssin.toml

VincentsSin:
Updated with most recent version.


VincentsSin:
I didn't notice that was open. I had only seen 2.0.0-hotfix was the latest in the index and made the PR out of instinct. Sorry about that!

PR #939, opened 2025-09-21
Add ESCHATOS manual

FlitPix:
Only now realized I could submit this here 😅

It's a very simple manual, there shouldn't be anything crazy going on but maybe I'm wrong


FlitPix:
no longer interested in implementing this

PR #947, opened 2025-09-25
Add Trails in the Sky the 3rd

nnguyen259:


nnguyen259:
Which of these are the actual blocker?

The 4 regions not connected and empty location group are all because it's more convenient to add them now than later when there are more stuffs to juggle within the apworld.

Still investigating the 2 more items than location but I have no idea why it keeps showing up.

And right now the client is the one handling the victory condition so I don't think it affects much?

nnguyen259:
ah ok I see, I'll hold off until the next release then.

Still trying to figure out why there are 2 extra items than location but no luck so far yet

index/tits_the_3rd.toml
```diff
@@ -0,0 +1,6 @@
+name = "Trails in the Sky the 3rd"
+home = "https://discord.com/channels/731205301247803413/1217595862872490065"
+default_url = "https://github.com/Archipelago-Trails-in-the-Sky-the-3rd/Archipelago-Trails-in-the-Sky-the-3rd/releases/download/{{version}}/tits_the_3rd.apworld"
+
+[versions]
+"0.1.5" = {}
```
Eijebong:
```suggestion
"0.1.6" = {}
```

PR #957, opened 2025-09-27
Update tboir.toml for the new version

AlexMagician:
the developer of this apworld has changed, and there is a new version v0.1.1 available, which is more feature-complete than the currently listed one (traps, deathlink, different checks and more customizability)


AlexMagician:
I've been notified that this implementation causes issues with the current AP setup, as mentioned in #923, therefore I'm closing this request.

PR #968, opened 2025-10-04
Add Lingo 2

hatkirby:


Eijebong:
@hatkirby You probably want to look at those generation failures since the rate is too high for what I'll accept here (looks all entrance rando related).

Also those timeouts are a bit worrying, I'm assuming it gets into swap hell when generated alone but the fact that it was able to hit a few of those when generated with a world that has 100 locations available in sphere 1 probably means that something is wrong. 10s to gen a single world is very very very very slow.

You can get the details from the zip files in the fuzz CI jobs. (and run the [fuzzer](https://github.com/Eijebong/Archipelago-fuzzer) locally if that helps. The "no-restrictive starts" job has one extra world per gen, from [this apworld](https://github.com/Eijebong/empty-apworld/releases/tag/0.0.1) which has basically no logic and 100 locations open in sphere 1.

hatkirby:
The non-restrictive-start timeouts are from yamls with every progression gating option disabled (which is not a normal use case), minimal accessibility, and entrance randomization enabled. I'm talking with some other devs right now and it looks like this is a bug in core GER that causes an infinite loop. I can raise an OptionError to prevent players from trying to generate a world with no progression items, but I also feel like no one is going to run into this on accident.

The other timeouts seem to be from solo yamls with all or almost all progression gating options enabled. Lingo 2 is notoriously complex, and does get into swap hell trying to make the early game work when generated alone. I've been working on adding more options to open the early game up even more but I think it will be difficult to completely eliminate this problem, especially considering how many players want to play solo worlds with maximum progression gating.

I can take a closer look at the ER related failures.

Eijebong:
Closing this as there's more info regarding failures in #1622 than in here

PR #994, opened 2025-10-25
Disable The Binding of Isaac Repentance

kingdragonite55:
The version in the index has been supplanted by a new version made by a different developer. Unfortunately that new version has poor fuzzing results and so can't be added here (#923). The old version should be disabled to avoid confusion.


Eijebong:
/taskcluster merge

Thx

PR #1047, opened 2025-10-29
Add Kirby Air Ride

nephalia:


DeDeDeK:
Awesome, thanks for the explanation! I'll look into fixing those. Next version 0.5.3 (or 0.6.0 possibly) will be the compliant one then.

PR #1058, opened 2025-11-01
Jigsaw: update to 0.9.1 (small bugfix)

spinerak:
This version is just a bugfix and you can use the same yaml as for 0.9.0!!!!! chrisWow


Eijebong:
A non YAML breaking update? Who are you and what have you done with spinerak?

PR #1065, opened 2025-11-04
Reenable Simpsons Hit And Run, Update to 0.4.3-b

Fore772:
Reenable Simpsons Hit And Run, Update to 0.3.6b


Fore772:
apworld dev very responsive
also may have PR'd the semver wrong. it needs a hyphen between version and letter, doesn't it?

Eijebong:
Doesn't load on linux 🙃

nmize1:
> Doesn't load on linux 🙃

I thought I'd fixed that. I'll have to track down what I missed. 

I'm not sure if this will pass fuzzing yet anyway. I have some options that purposefully fail gen because they're not implemented yet and I haven't gone back to make them fuzzer friendly. I only heard of this a few days ago.

Eijebong:
I think the issue is the `\` in the file paths when you try to load it

PR #1068, opened 2025-11-04
Add Sly 2 (ChrisM Event Ver)

Awsomeman328:
This will attempt to use a specific & special build version of this APWorld that guarantees that more locations than items will be generated for itself. Don't use the regular release version from the official repo until this has been officially addressed there.

### DIFF
--- a/index/sly2.toml
+++ b/index/sly2.toml
@@ -6,5 +6,5 @@ default_url = "https://github.com/Awsomeman328/APSly2/releases/download/v{{versi
 # the regular public release builds (currently) have a failure rate of about 80% with the fuzzer tool and shouldn't be used until fixed.
 
 [versions]
-"0.1.3-chrism" = {}
+"0.1.3-chrism" = { }
 


Mysteryem:
Have you asked the developer of the apworld about distributing this modified version of the apworld?

Awsomeman328:
Not directly, but elements of these changes seemed to be undesirable to them and others in the thread, mainly lowering the max number of "Clockwerk Parts" Items from 100 to 50, which are items specifically unique to AP and not from the base game so they don't have a regular location to pull them from.

There are plans to program in more locations to handle the lofty extra item count, and ideally a smarter validation system to estimate aprox. item and location counts and to adjust the user's settings to better avoid the "too many items/not enough locations" error, but there wasn't enough time for either to submit it in time for the Chrism Big AP Event. Plus it'd be weird to enact this restriction on the APWorld for the official release just to remove it again once those extra locations get added.

I guess it is also weird to build a modified apworld release just for a single event, but I figured doing this would be more preferable.

Awsomeman328:
Personally I was seeing this more of how like a manual is an individualized apworld for a player, but I understand where you're coming from and I hadn't quite thought of it that way exactly. I'll clean my mess out of this branch and shelve this game until we get a proper official release to pass the fuzzer.

PR #1070, opened 2025-11-05
Update spelunky2.toml

DDR-Khat:


Mysteryem:
There is likely a missing indirect condition for the `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` entrance when the Easy goal is enabled.

The rule for `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` checks `can_access_sunken_city()`, which checks `can_obtain_qilin()`, which checks for access to the `DUAT` region, `ABZU` region and the `EXCALIBUR` location (which checks for access to the `EXCALIBUR` location's `.parent_region` `WorldName.TIDE_POOL`).

The `DUAT` and `ABZU` regions are registered when the Easy goal is enabled, but the `WorldName.TIDE_POOL` region is missing. If access to `WorldName.TIDE_POOL` *could* be the last requirement to satisfy to gain access to the `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` entrance, then an indirect condition is needed to ensure correct accessibility and deterministic generation with a fixed seed.

---

There is also a `SUNKEN_CITY_TO_COSMIC_OCEAN` + `WorldName.SUNKEN_CITY` indirect condition that does not seem to be necessary any more. The old entrance rule checked for access to the `Sunken City` region, but the new entrance rule checks `can_access_cosmic_ocean()`, which only checks for having items, and not any access to regions.

A single extraneous indirect condition won't cause any generation issues, it'll only make generation the tiniest bit slower than needed, from doing unnecessary work.

DDR-Khat:
> There is likely a missing indirect condition for the `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` entrance when the Easy goal is enabled.
> 
> The rule for `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` checks `can_access_sunken_city()`, which checks `can_obtain_qilin()`, which checks for access to the `DUAT` region, `ABZU` region and the `EXCALIBUR` location (which checks for access to the `EXCALIBUR` location's `.parent_region` `WorldName.TIDE_POOL`).
> 

If the player's goal is EASY, they do not go to Sunken or Cosmic Ocean; Both areas are excluded from having items added if that is their difficulty setting.

> The `DUAT` and `ABZU` regions are registered when the Easy goal is enabled, but the `WorldName.TIDE_POOL` region is missing. If access to `WorldName.TIDE_POOL` _could_ be the last requirement to satisfy to gain access to the `RuleNames.NEO_BABYLON_TO_SUNKEN_CITY` entrance, then an indirect condition is needed to ensure correct accessibility and deterministic generation with a fixed seed.
> 
Under normal gameplay;
The player must receive the "Tablet of Destiny" from that area, in order to correctly pick the right "Ushabti" which turns into "Qilin" which is used to ascend NeoBabylon-4 into SukenCity . 
The correct "Ushabti" is always randomized (it -CAN- be forced but then this opens up brute-forcing and breaking logic)

If the player has learned skips, there are alternate ways the player can ascend to Sunken CIty without that quest line.

> There is also a `SUNKEN_CITY_TO_COSMIC_OCEAN` + `WorldName.SUNKEN_CITY` indirect condition that does not seem to be necessary any more. The old entrance rule checked for access to the `Sunken City` region, but the new entrance rule checks `can_access_cosmic_ocean()`, which only checks for having items, and not any access to regions.
> 
> A single extraneous indirect condition won't cause any generation issues, it'll only make generation the tiniest bit slower than needed, from doing unnecessary work.

That's a good shout and I need to fix that, as the player MUST be able to get into Sunken City in addition to items.

DDR-Khat:
> Numbers are free, I promise you won't run out of them :p
> 
> Also that test means that there's 2 bugs to fix now

I know that :P I just don't like issuing an update for one tiny thing. But I will see what can be achieved if I get time tomorrow, I've been wanting to get the client issues resolved anyway but again time is the limiting factor.

PR #1088, opened 2025-11-08
Downgrade Sims 4 APWorld to properly stable release

benny-dreamly:
#1009 updated the APWorld to a version that is **not considered stable enough** to use in public syncs or asyncs without prior knowledge from the host. There are several large client issues that are still in need of fixing, along with multiple other things that are probably unknown at this point. 

This is currently the best course of action until I have time to finish the beta.

A future PR will add the properly stable version of the APWorld that is recommended for public use.


benny-dreamly:
> Why not just roll back to 1.6.1 until 2.0 is ready?

I'm planning on doing that in another PR but I can just as easily do that here 

(1.7.4 is also fine too)

benny-dreamly:
> Don't make extra work for Eijebong by doing two separate PRs 😛

good point lol

PR #1103, opened 2025-11-13
Add Xenoblade Chronicles manual

Garfunkiel:
I have modified and am submitting (with permission) this manual APWorld created by Xanderoni (https://discord.com/channels/1097532591650910289/1216897393019322388).  

Local testing only shows errors relating to pickling options (and I have provided a meta toml to ignore these, matching other manual APWorlds in this index).  Fuzzing has a 0% failure and 0% timeout rate on my machine (10000 runs), though keeping the timeouts to 0% requires either the collectopaediasanity option to be meta'ed to false or a slightly higher timeout (-t 20).  I've improved the rate for that option from what it used to be (it needed -t 60 before my changes), but please let me know if it's still slower than whatever threshold you're willing to allow into the index or if there are other problems I need to fix.  Thanks!


index/manual_xenobladechroniclesv2_xanderoni_skipperki1.toml
```diff
@@ -0,0 +1,7 @@
+name = "Manual_XenobladeChroniclesv2_Xanderoni_skipperki1"
```
Eijebong: This doesn't load because the name of the apworld must match the name of the folder within it. `manual_xenobladechroniclesv2_xanderoni_skipperki1` vs `manual_xenobladechroniclesv2_xanderoni`

meta/manual_xenobladechroniclesv2_xanderoni_skipperki1.toml
```diff
@@ -0,0 +1,95 @@
+[["test.general.test_options.TestOptions.test_pickle_dumps (game='Manual_XenobladeChroniclesv2_Xanderoni_skipperki1', option='goal')"]]
```
Eijebong: Can you remove those, they're outdated anyway since that test name has changed? I'll generate them through CI automatically

index/manual_xenobladechroniclesv2_xanderoni_skipperki1.toml
Garfunkiel: Done.

Garfunkiel:
I think I've resolved all outstanding issues with it, but please let me know if any other changes are needed.  Thanks!

Garfunkiel:
I'll see what I can do to make it faster.  What factor, roughly, would you say it needs to be improved by?  Just looking for some idea of what threshold or target you're comparing it to.

PR #1108, opened 2025-11-15
Update Hitman WOA to 0.6.0

kingdragonite55:


Eijebong:
Love it when the diff doesn't even load... Really bodes well... What the heck is in there

Eijebong:
Probably hitting this... But a lot worse... https://github.com/Eijebong/Archipelago-lobby/issues/95

PR #1114, opened 2025-11-26
Update APworld version of Crystal Project to 0.11.2

OriginalTomPouce:
I see there is another PR of Crystal Project from 0.9 to 0.10.

0.11.2 has a lot of bugfixes, and a big chunck of logic revamp.
Hope it passes the fuzzer !

### DIFF
--- a/index/crystal_project.toml
+++ b/index/crystal_project.toml
@@ -5,4 +5,4 @@ default_url = "https://github.com/Emerassi/CrystalProjectAPWorld/releases/downlo
 [versions]
 "0.7.0" = {}
 "0.9.0" = {}
-"0.11.2" = {}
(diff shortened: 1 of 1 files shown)


Eijebong:
/taskcluster r+

```
            if self.options.max_level.value > potential_max_level:
                raise Exception(f"For player {self.player_name}: yaml settings were too restrictive. Needed at least {-(self.options.max_level.value // -potential_progressive_level_size)} Progressive Levels, but only room for {max_progressive_levels} Progressive Levels in the pool. "
                                f"This is usually caused by mods that add more items than locations. Change settings and regenerate.")
```

This should be an `OptionError` and it still struggles a bit with restrictive starts. But that look ok enough.

PR #1118, opened 2025-11-29

nephalia:
this time FOR SURE


nephalia:
no changes to logic but i'm updating the toml anyway chrisOoga

PR #1123, opened 2025-12-01
Update pokemon_bw.toml

Radis7Noir:


Eijebong:
/taskcluster test-fuzz

Actually nvm, looks like it a very dumb fix, sorry for the spam though

PR #1147, opened 2025-12-17
Add Portal 2

fewffwa:


Eijebong:
I'm all for ignore restrictive starts but that it a bit much... Basically can't generate on its own

Eijebong:
Was expecting that to drop down to 0 so there's probably still something wrong 🤔

PR #1151, opened 2025-12-17
Add Hacknet

nephalia:


Eijebong:
/taskcluster test-fuzz

Just out of curiosity, no r+ cause of direct random usage

PR #1153, opened 2025-12-18
Add We Love Katamari Reroll

sapphisylph:
i think everything should be good with this? if it fails ill look at what could be fixed if it's not a fundamental Game issue


index/manual_welovekatamarireroll_sapphisylph.toml
```diff
@@ -0,0 +1,7 @@
+name = "Manual_WeLoveKatamariReroll_sapphisylph"
+display_name = "Manual: We Love Katamari Reroll"
+home = "https://discord.com/channels/1097532591650910289/1348016489181085776"
+default_url = "https://github.com/sapphisylph/We-Love-Katamari-Manual-APWorld/releases/download/v{{version}}/manual_welovekatamarireroll_sapphisylph.apworld
```
Eijebong: Missing a double quote here.

Eijebong:
This is **way** too slow holy moly... I don't know how it consistently stayed below the timeout threshold but 26mn for 5k gens is not acceptable

PR #1156, opened 2025-12-19
Update Lego Star Wars: The Complete Saga to v1.2.1

Mysteryem:


Mysteryem:
I'll fix one of the options putting too many of an item in starting inventory tomorrow, it's not a big problem, just mildly annoying to have some locations that might require some currency grinding coming into logic earlier than people might expect.

PR #1175, opened 2026-01-01
Update pokepark.toml

ddrox13:
Add version 1.1.0. Hopefully this works?

### DIFF
--- a/index/pokepark.toml
+++ b/index/pokepark.toml
@@ -6,4 +6,4 @@ default_url = "https://github.com/Mekurushi/Archipelago_Pokepark/releases/downlo
 [versions]
 "1.0.1" = {}
 "1.0.3" = {}
-"1.1.0" = {}
+"1.1.1" = {}


Eijebong:
Hrmf, that is technically below 1% but that is quite the regression and it looks like an actual logic bug that is just getting hidden by the empty apworld.

@Mekurushi FYI, looks like you introduced some logic issue between 1.0.3 and 1.1.0 that's getting you from a 0% failure rate to almost 4% when using https://github.com/Eijebong/Archipelago-fuzzer to generate with random YAMLs.

Mekurushi:
oh that's the entrance rando. I know of the higher failure rate with the entrance rando, because I implemented just a basic prototype with very simple logic. The failure rate was acceptable for me, because it only affects the new experimental feature.
Truth be told I didn't even know of the existence of this index and the fuzzer failure rate requirement. I can focus on improving entrance rando and getting the failure rate down for the next update.

Mekurushi:
@Eijebong I released right now 1.1.1 that should fix the high failure rate from my perspective. At least locally fuzzer doesn't have any failures anymore. Can we update the index for that version then?

PR #1183, opened 2026-01-05
Un-yeet Cartoon Network Punch Time Explosion XL

BunkBunker:
Hopefully this should fix all the issues my manual was having


BunkBunker:
Figured I'd delete the apworld that was still in there too since it'd still have issues anyway

PR #1184, opened 2026-01-05
Reactivating Skyward Sword with 0.5.4

harmjan387:


Mysteryem:
I don't know why `eval` is being used. It looks to me that the developers could instead generate a python file with all the rules in it, avoiding use of `eval` entirely. The Batreux rules would need to be handled slightly differently by making them take a count argument.

Almost no benefits of compiling rules at runtime seem to be used.
- Non-literal int/str/tuple/list are not compiled into literals directly in the code (except for `batreaux_requirements` rules)
- Checking a player's options is not replaced with the result of checking those options
- Extra function calls like `state.has()` are not inlined

PR #1192, opened 2026-01-07
Update banjo_tooie.toml to v4.11.4

tommadness:


Eijebong:
Yeah... Still need a PhD to make a valid YAML :/

tommadness:
> Yeah... Still need a PhD to make a valid YAML :/

I assume it's OptionsErroring because of the restrictions on Move Rando without Nest/Signsanity, logic_type restrictions, etc?

I can provide some example yamls if needed. I *might* be able to throw together some kind of trigger block but no guarantees.

Eijebong:
/taskcluster r+

Yeah, I fuzzed it manually with random valid YAMLs I had around  from previous community syncs, looks fine

PR #1193, opened 2026-01-07
Add SIS Support

Typ0Codes:
Adds my unreleased game Shipped in space. it will be releasing before summer. i would like to have it on bananium if possible before the game releases so i can do stuff like play it in big asyncs or community APs. 

if you need a build of the game for testing purposes i can get you one. 

ok thank you!


Eijebong:
Looks like this contains no logic whatsoever? You can just walk into the goal from start?

Means that if your game take 3h to complete because you have to go through all your levels and they have their first progression item in your sphere 1 (which is everything in this case), you'll just wall them entirely for 3h. I do try to avoid apworlds like that that are "unfriendly" (for the lack of a better word) to multiworlds because of their lack of logic

Typ0Codes:
> Looks like this contains no logic whatsoever? You can just walk into the goal from start?
> 
> Means that if your game take 3h to complete because you have to go through all your levels and they have their first progression item in your sphere 1 (which is everything in this case), you'll just wall them entirely for 3h. I do try to avoid apworlds like that that are "unfriendly" (for the lack of a better word) to multiworlds because of their lack of logic

the way the game works is you go level by level. there are no required items for completion only items that make it easier to progress. the game usually takes about an hour in archipeligo shorter if you get good items.

Typ0Codes:
if i where to add logic it would just be connecting one level to the next. the game is meant (in archipeligo) to be a side game where you grab a few checks while BKed in your main game. 

if there is anything i could add to make it more "freindly" i wouldnt be opposed.

Typ0Codes:
OK i updated it with the keys for every 10 levels. having played it a bit solo i think this was a really good idea and it will be a way more fun AP.

thanks for the suggestions!

Eijebong:
Well you got keys but I'm still not seeing any logic saying that you need said keys to access levels

Typ0Codes:
> Well you got keys but I'm still not seeing any logic saying that you need said keys to access levels

There is? 

image

Eijebong:
Look at the diff? :) I don't think it is in your 2.0.0 version

Typ0Codes:
> Look at the diff? :) I don't think it is in your 2.0.0 version

ok the release has been updated

Typ0Codes:
> This is 2.0.0 https://apdiff.bananium.fr/tests/Ur5cIT5rRqaLWohEQ15hiQ

im guessing its because i do this in locations.py

`def place_locked_items(world: ShippedInSpaceWorld) -> None:

    Lvl40 = world.get_location("win")

    Lvl40.place_locked_item(world.create_item("Winstate"))`
    
im pretty sure i can get this to work without it if i just do it in client.

Typ0Codes:
never mind that seems like a bad idea.

Typ0Codes:
hey i want to try to do this for the chrism ap next week. what are the odds we could get this in before then?

PR #1195, opened 2026-01-08
Update gl.toml

Awwa111:
Update Gauntlet Legends to latest version.

Adds better unlimited speed rather than simply a long duration, randomization of portals to different levels, and defeat X bosses to win instead of just the final boss.


Awwa111:
Oh woops, I just saw a fancy button and pressed it, and it started doing tests so I was like.... I won't touch this anymore

PR #1200, opened 2026-01-09
Add Plants vs Zombies: Replanted

nephalia:


Eijebong:
/taskcluster test-fuzz

Let's see how it goes, but .git in the apworld means nope anyway

PR #1208, opened 2026-01-09
Add Sly 2

NikolajDanger:


Eijebong:
/taskcluster r+

Guess I'll take it because I don't want to be too annoying about apworld sizes, thank you for making it at least more reasonable. (really wish we had a better solution for UT to get those images rather than having them baked in the apworld)

PR #1261, opened 2026-01-12
Update manual_residentevil4_vincentssin.toml

VincentsSin:
Minor update to correct win-con errors, plus other minor logic corrections


Eijebong:
Given how slow this is I cancelled the other fuzzing task because the no-restrictive one is not very happy

> [task 2026-01-12T19:02:53.374+00:00] 1400 / 5000 done. 562 failures, 11 timeouts, 0 ignored.

Eijebong:
Then something else is wrong

PR #1267, opened 2026-01-12
Update Outer Wilds to 0.3.20

palex00:


palex00:
I have checked: These errors also occur on the current 0.3.19 live version.

palex00:
I have reported the failures to the dev. Ixrec was shocked but is appreciative and is investigating. They cannot guarantee a fix if it's an obscure combination that players never use.

Considering this is *not* a regression from 0.3.19 which is on the index, I'd still advocate for the update to go through, despite fuzzer failures.

Eijebong:
Ah it's probably https://github.com/Eijebong/Archipelago-fuzzer/commit/9502a8f639a914807d61e1c6ae5fe29f600bdcad

PR #1277, opened 2026-01-13
Add Manual: Ace Combat 7

nephalia:


Qipanjin:
Sadness. Well was worth a shot trying to get it added I guess. Even though I understand none of the coding garble as to why it failed.

PR #1282, opened 2026-01-13
Add Nodebuster

Eijebong:


palex00:
Nodebuster has barely any logic. In an Async I hosted a player reported over 200 locations were Sphere 1. And they were impossible to get with "nothing" as they definitely required health + stuff. I do not recommend adding this to the index. *And that's coming from me*

PR #1293, opened 2026-01-14
Add Garfield Kart - Furious Racing

robotzurg:


robotzurg:
> `[task 2026-01-14T18:57:29.689+00:00] Exception: Apworld must contain a directory named 'garfieldkart'. Found: ['garfkart']`
> 
> The toml file must have the same name as the apworld one (so `garfkart.toml`)

Fixed!

Eijebong:
/taskcluster r++

No rush on the test failure, it's not really relevant (well, it is but I don't care much, it won't break anything). Might want to look at the

PR #1294, opened 2026-01-14
Update GLYPHS to 1.1.1

BuffYoda21:


Eijebong:
That's what I call a regression

BuffYoda21:
turns out the issue came from if the user decided to enable traps but also not specify any traps 😅

PR #1297, opened 2026-01-14
Add Gunfire Reborn manual

Eijebong:


Eijebong:
/taskcluster r++

Dev said they'd fix the issue but they happen early enough to get caught by the YAML checker and in a very clear case that I'm not worried about it.

PR #1312, opened 2026-01-17
Add Isles of Sea and Sky

kingdragonite55:
Creator's Github versioning does not adhere to semver


gerbiljames (changes requested):
@kingdragonite55 Name has to match the apworld exactly, I missed this one too to be fair

PR #1313, opened 2026-01-17
Add Oblivion Remastered

JimmySplodge:


index/oblivion.toml
```diff
@@ -0,0 +1,6 @@
+name = "Oblivion Remastered"
+home = "https://discord.com/channels/1085716850370957462/1379282061071749221"
+tags = ["ad"]
+
+[versions]
+"0.4.6" = { url = "https://github.com/POD-io/Oblivion-ArchipelagoWorld/releases/tag/v0.4.6" }
```
Eijebong: This needs to point to the apworld download itself.
Also if the version is actually 0.4.6 you can use the default_url pattern to make future updates less painful

PR #1318, opened 2026-01-18
Update tboir

NaveTK:
Update to new maintained version, also isaac got its own channel as the future-game-design thread got close to 500 members


NaveTK:
> Also don't know if you need transparency on those images but you could cut the size of the apworld in half by converting those images to jpg

I need it on some but not all, can probably optimize apworld size a bit in the next version. But a bigger size is more or less inevitable with the map tracker bundled into the client.

PR #1319, opened 2026-01-18
Add Manual_Touhou Fumo Racing_AwesomeZuils

zuils:


zuils:
should fix that rare gen issue now

PR #1323, opened 2026-01-18

Foopyo:


Eijebong:
Dupe of #193  #1118 and we know that the failures blocking this from getting into the index are not fixed by the thing mentioned in the release notes

PR #473, opened 2025-05-08
Create turnipboy.toml

Fore772:
Added turnipboy.toml

PR #1325, opened 2026-01-18
Add Turnip Boy

gerbiljames:


Eijebong:
https://github.com/Eijebong/Archipelago-index/pull/473#issuecomment-2862594796

gerbiljames:
> [#473 (comment)](https://github.com/Eijebong/Archipelago-index/pull/473#issuecomment-2862594796)

Looks like its due to [this](https://github.com/ArchipelagoMW/Archipelago/issues/4224), all of the failures have a mix of minimal and full accessibility and changing them to match makes the failures go away

Gingerbit25:
I was looking at the Index and noticed that Turnipboy is disable due to a bug. But honestly, I don't know if it refers to an actual bug or the fact that the player gets BK for a long amount of hours before it can go for more checks in long syncs
From my experience, that's normal

PR #1334, opened 2026-01-19
Update ladx_beta.toml

palex00:


palex00:
> I mean... It was already disabled :p

Disregard, I am an idiot. Anyway, can you still fuzz / run assertions? </3 maybe I can get some of the issues knocked out

PR #1343, opened 2026-01-20
Add Rabbit and Steel

Eijebong:


Mysteryem:
Nondeterministic generation in picking starting class in `generate_early` (World.py lines 135 to 137)
```python
        if self.options.class_sanity:
            self.starting_class_name = self.random.choice(tuple(set(class_names) - set(self.options.exclude_class)))
            self.multiworld.push_precollected(self.create_item_with_type(self.starting_class_name, "Classes"))
```
Sets are unordered, so the tuple created from the sets will have a nondeterministic order.

PR #1351, opened 2026-01-21
Update osu.toml

palex00:
The question is also if we should keep the old betas since it's the same link.


Eijebong:
no lol, how did this happen... Yeet that -beta one please

Eijebong:
This must be hitting another pathological case in the diff viewer because that's literally not loading :/

Eijebong:
Doesn't gen its own template

PR #1352, opened 2026-01-21
Update R.E.P.O. to 0.3.0

Kafuruu:
Adds compatibility for the latest/monster update for REPO. Also fixes lag with larger syncs by adding a config to disable multiworld messages on the truck screen.


index/repo.toml
```diff
@@ -1,6 +1,6 @@
-name = "R.E.P.O"
+name = "R.E.P.O."
```
Eijebong: This needs to match the name of the world, you cannot change it like this

Kafuruu: ah my bad

PR #1362, opened 2026-01-23
Add Overwatch manual for community sync.

LexHDev:


index/manual_overwatch.toml
```diff
@@ -0,0 +1,6 @@
+name = "Manual_Overwatch2_Fafale"
+display_name = "Manual: Overwatch 2"
+home = "https://github.com/Fafale/manual_overwatch2_fafale/releases/latest"
+
+[versions]
+"v1.1.1" = {local = "../apworlds/manual_overwatch2_fafale.apworld" }
```
Eijebong: Oh and the version needs to drop that `v`, sorry I had missed that

Eijebong:
Not sure why you're not using the apworld in the github release. I would much prefer that over adding the apworld directly to the repo (that should be limited to apworlds that are not downloadable from anywhere else). Also that apworld contains a full .git directory which is a nope, that should not be distributed.

PR #1391, opened 2026-01-30
Update Resident Evil 2 Remake to 0.3.3

AmondoVagyok:


Eijebong:
/taskcluster test-fuzz

From a quick read I feel like this is going to be worse than before with traps not being replaced by items forced on locations

Eijebong:
Ah another "can't fail fuzzing if you can't fuzz without 100% option errors". Given failures are `list index out of range` I'm not going to take that, that's a bug that should be addressed

AmondoVagyok:
@FuzzyGamesOn Sure, but just a heads up, it may or may not get accepted with a modified release artifact.
Another game did it before https://github.com/Eijebong/Archipelago-index/pull/1373#issuecomment-3807245928

Eijebong:
/taskcluster test-fuzz


> it may or may not get accepted with a modified release artifact.

I wouldn't refuse it just because of that but it'd be nicer if versions actually meant something instead of "which 0.3.2 do you have". (+ if stuff is already in the index it'd be a "was that change real or a malicious replacement")

AmondoVagyok:
> /taskcluster test-fuzz
> 
> > it may or may not get accepted with a modified release artifact.
> 
> I wouldn't refuse it just because of that but it'd be nicer if versions actually meant something instead of "which 0.3.2 do you have". (+ if stuff is already in the index it'd be a "was that change real or a malicious replacement")

I totally understad you point, I too prefer a smaller release instead, just to make changes clear.

AmondoVagyok:
I would like to say my bit on this as the one starting this PR in the first place. I would like to apologise for both parties here, because I truly didn't mean to start any drama. I had no ill intentions towards either party, and I'm truly sorry it came to this.

PR #1393, opened 2026-01-30
Update BitsAndBops to 1.0.6

xMcacutt:


Eijebong:
That's a bit of a problem, means the game is always beatable from the start.

Eijebong:
/taskcluster fuzz -r 5000 -n 1 --dump-ignored

Eijebong:
There we go. Just need that completion condition fixed now :p

Eijebong:
/taskcluster test-fuzz

Nope, we can reuse PRs, all CI is always based on the state of your branch at the time it's run so it doesn't matter.

That manifest removal looks like a mistake but I don't think it was fully valid anyway? @gerbiljames if you want to have a look. (not a blocker for me right now but something to consider for the future)

Eijebong:
/taskcluster r+

Thx for fixing :)

PR #1404, opened 2026-01-31
Adds Hi-Fi RUSH by Trev (TRPG)

tommadness:
Adds the non-manual Hi-Fi RUSH


tustin2121:
> It's not... I'm just asking if you still needed the manual if there was a proper implementation, I'm fine with having both (platinum has both for example).

Fair enough. My apologies. I would like if both were available, if that's okay.

PR #1407, opened 2026-02-02
Add version 0.3.1 for TMC

eternalcode0:


eternalcode0:
> I feel like that `OptionError` should always be thrown, submitting a YAML that submits more progression than non excluded locations has big main character energy and IMO should just be slapped down. Also if every world was allowing this, we'd never get anywhere. This is an improvement over not saying anything ever but I'm not sure how I feel about it just shoving user errors under the rug

@Eijebong Now that you mention it I do agree that this should always throw regardless of player count. Just for background though this should only occur from people making some absolutely absurd yamls (that the fuzzer could easily create) that include at least 3/5 of the max mcguffin hunt, *no* extra location shuffle, but excluding non element dungeons. Those are kinda contradictory, someone using those settings is probably confused about what kind of run they want.

Regardless I do want to improve this error in the future. We just had a critical bug that could corrupt saves discovered in v0.3.0 so I wanted to get a fix for the save corruption bug out *first*. This error message was just meant to be temporary until I figure out a better solution to tally the items/locations the yaml adds.

PR #1409, opened 2026-02-02
Add Ty the Tasmanian Tiger 2

xMcacutt:


Eijebong:
@xMcacutt Can you check those location/item count failures? Looks like you have settings submitting more items than locations to the multiworld.

PR #1423, opened 2026-02-07
Update three home URLs

kingdragonite55:
Added a home URL to Hitman WoA, and updated the URLs for Kirby Super Star and Outer Wilds since they both got their own channels.


Eijebong:
/taskcluster r+

Thanks!

PR #1424, opened 2026-02-08
Update Portal 2 to 0.5.6

Radis7Noir:


GlassToadstool:
0.5.5 out, if this doesn't work I'm giving up programming forever

PR #1425, opened 2026-02-08
Update peaks of yore

Eijebong:


Eijebong:
The state leak is reversed, it removes logic, reported here: https://discord.com/channels/731205301247803413/1332506808182378677/1470075746184855573

PR #1428, opened 2026-02-08
Added Void Sols toml

cookie966507:


cookie966507:
@gerbiljames good callout on the test case. This was actually one of the last bugs I fixed, but I never updated the test case for it. I can take care of that. Thanks for the other fixes too. This is my first AP world so I'm surprised this was all that was wrong lol

Eijebong:
Well keep in mind we don't review for code quality here :joy: It doesn't have something that jumped to James wrt security, it doesn't have outstanding known bugs and it gens. That's all we check for. A passing mark here doesn't mean a proper code review won't raise questions and other potential problems.

PR #1435, opened 2026-02-10
Update Void Sols toml

cookie966507:
Released a new version for one found bug relating to the Gatekeeper access rule


cookie966507:
Hm I'm wondering why that options check still failed. I wonder if something didn'tt update correctly.

Eijebong:
Nah it says "unexpected success". When we accept worlds that fail test we note those down so that we can keep track of what failed or got fixed. That's all. It's clean now

PR #1464, opened 2026-02-16
Update Rain World to 1.5.2

Radis7Noir:


Eijebong:
That's a regression

SaltiestSyrup:
Hi this is world dev 👋 

I've seen this and am working on fixing these issues for 1.5.1. They didn't pop up in my own fuzzing because I was mistakenly using `--skip-output` on my main run :/

The determinism check is new to me and apparently has been an issue at least since the last version (1.4.0) - I need to do further testing to see if it goes further back but thought I should make that known.

SaltiestSyrup:
Status on this?

Eijebong:
/taskcluster r+

Really not a fan of these "I'm the main character" settings. 🤷
Looks like it shouldn't be causing issues in big worlds though.

Something worth noting, while looking at it myself, you have `prioritizable_passages` which is a list of passages that should be prioritized in that case but it's never used (it's imported in `__init__.py`). I don't know if it's on purpose or a mistake but you might want to have a look.

SaltiestSyrup:
I kinda get what you mean but really it's just "I want to make some of these priority but random please", not any different than prioritizing them yourself.

Huh, I had literally no idea that list was there. I wasn't the one who implemented this option, I guess they made the list and just forgot to utilize it? I'll do something with that, thanks for letting me know.

PR #1465, opened 2026-02-16
Update Buckshot Roulette to 0.1.1

Radis7Noir:


Eijebong:
/taskcluster r+

Not really fond of the check for 1 player in the multiworld to raise OptionErrors but given there was no failure on n2 it's probably maybe fine...

PR #1468, opened 2026-02-16
Update Final Fantasy IV: Free Enterprise to Version 7

palex00:
Was disabled due to 25% fuzz failures. Version 7 makes it 0.9% fuzz failure.


Eijebong:
Nice 50/50  on check-item-location-count

PR #1489, opened 2026-02-19
Update Plants vs. Zombies Replanted to v1.5.0

AmondoVagyok:


AmondoVagyok:
It'll most likely have an update again, which will have AP fixes (manifest has incorrect formatting in it) so I'll close this one

PR #1495, opened 2026-02-21
Update Plants vs. Zombies Replanted

AmondoVagyok:


AmondoVagyok:
> /taskcluster r+
> 
> 2 out of those 5 failures are accessibility issues, might want to report/investigate that

Will do, thank you!

PR #1523, opened 2026-02-24
Update crystal_project

Eijebong:


Eijebong:
Too much stuff that really shouldn't be in there...

PR #1535, opened 2026-02-24
Update th06

Eijebong:


Eijebong:
Solo gen failures seems to be because sphere 1 is very tiny (3 locs) and requires a lot of items to get out of.

PR #1563, opened 2026-02-26
Add Manual: TBOI: Four Souls

nephalia:


Eijebong:
Doesn't genenrate its own template `[task 2026-02-26T22:29:55.330+00:00] KeyError: "Duplicate key false found in YAML. Already found keys: {'true', 'false'}."`


It produces:

```
death_link:
  'false': 0
  'true': 0
  'false': 50
```

Fabby29:
Updated version of the APWorld, it should generate its own template now.
[manual_isaacfoursouls_fabby29.zip](https://github.com/user-attachments/files/26742229/manual_isaacfoursouls_fabby29.zip)

PR #1601, opened 2026-03-06
adding Final Fantasy X manual

KuroYukiHane1:


index/Manual_FinalFantasyX_KuroYukiHane.toml
```diff
@@ -0,0 +1,5 @@
+name = "Manual_FinalFantasyX_KuroYukiHane"
+display_name = "Manual: Final Fantasy X"
+
+[versions]
+"0.0.3" = { local = "../apworlds/manual_FinalFantasyX_KuroYukiHane-0.0.3.apworld" }
```
Eijebong: This does not match the file name in the repo so that won't work

KuroYukiHane1:
ah it's case sensitive? hopefully i haven't missed anything this time

Eijebong:
Nope. The `file = ` field just needs to point to something existing (which it doesn't anymore). The toml file needs to be named the same as what the apworld is named to work.

KuroYukiHane1:
ah it's case sensitive? hopefully i haven't missed anything this time

> Nope. The `file = ` field just needs to point to something existing (which it doesn't anymore). The toml file needs to be named the same as what the apworld is named to work.

does that include the version exstention? I didn't see any tomls in the repo with that exstention so i didn't include it. I've updated the toml name to be all lower case now so now the toml, the apworld, and the apworld reference in the toml are all lower case. 
Is that all or are you saying i'm missing another file entirely?

KuroYukiHane1:
Do I need to do anything else here? I see it marked as waiting for author but i don't see anything to do. do I need to confirm something on the review changes link?

PR #1622, opened 2026-03-11
Added Lingo 2

AshCauseYes:


Eijebong:
/taskcluster test-fuzz

Previously tested in #968

hatkirby:
It looks like a significant number of these errors are due to incorrect filtering of non-ASCII characters for Custom Mint Ending. This PR also contains a local apworld but I think it would be better to point at my website like the previous PR did. I might reopen my PR after fixing the Custom Mint Ending issue. I suspect however that some timeouts are still inevitable due to how complex Lingo 2 is and how restrictive the early game is when using Progressive Letter Shuffle.

Eijebong:
Yeah. It'd be preferable for sure. Feel free to reopen your PR or open a new one. I don't see any timeouts in this version so I'm not sure if they're actually going to be an issue?

PR #1638, opened 2026-03-19
Update pokemon_bw.toml with new version

mzweiten:
Add version '0.3.24' to the versions list

### DIFF
--- a/index/pokemon_bw.toml
+++ b/index/pokemon_bw.toml
@@ -4,6 +4,3 @@ default_url = "https://github.com/BlastSlimey/PokemonBWAP/releases/download/{{ve
 
 [versions]
 "0.3.27" = {}
-"0.3.17" = {}
-"0.3.14" = {}
-"0.3.13" = {}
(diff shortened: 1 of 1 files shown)


Mysteryem:
Both timeouts from the default fuzz are infinite loops in `generate_early` in `generate.wild.generate_wild_encounters()`.

`logic_slots` sometimes runs out of elements before `species_checklist[0]` is exhausted, so `while len(species_checklist[0]) > 0:` loops forever.

The code that I believe to be the cause is `generate.encounter.checklist.add_species_to_check()` because this is what causes the length of `checklist[0]` to grow larger than `logic_slots`.

```python
def add_species_to_check(checklist: tuple[list[str], set[str]], species: str) -> None:
    if species not in checklist[1]:
        checklist[0].append(species)
```

I suspect this should instead be checking that `species` is in neither `checklist[0]` (to be checked species) nor `checklist[1]` (already checked species), so it currently can result in species being added to `checklist[0]` twice, and eventually making `checklist[0]` larger than the remaining `logic_slots`.

What confuses me, is that this code was refactored 6 months ago on the `main` branch, and makes the change I suspected was needed: https://github.com/SparkyDaDoggo/Archipelago/blob/31352008dadaf03ad28a230cc35c659ebda90630/worlds/pokemon_bw/generate/__init__.py#L85-L90, but this refactored code, despite being 6 months old, isn't in any release? With current development instead taking place on the `pokemon-bw-0-3-legacy` branch where the problematic `add_species_to_check` is older (last updated 7 months ago according to the git blame).

PR #1661, opened 2026-03-28
Update Devil May Cry 3 to 0.5.0

AshIndigo:
Adds an entry for 0.5.0, I think I did this right?


Eijebong:
> check-collect-accessibility 

This should not fail. It means that collecting an item can reduce what locations are accessible.

AshIndigo:
Yeah, been trying to figure out a small bit of non-determinism before I submit 0.5.1 for review. 

`has_group_unique` looks to be what I need though for the essences, as well as another rule involving fragments. Thanks for pointing it out.

PR #1662, opened 2026-03-28
Update Metroid Fusion to 0.0.21

kingdragonite55:
The original creator of the apworld has stepped down, there is a new group of people working on it


Eijebong:
> AssertionError: Sector 6 Crossroads already exists in region cache.

Doesn't even gen its own template

kingdragonite55:
We love updating release artifacts 🙃 At least jsonschema finally got removed

Eijebong:
/taskcluster test-fuzz

Bit worried about that Bounce in the main loop (it's sending the current map twice a second, at least scoped to itself but I think that's what Set/SetNotify is for?).
Also I don't understand what happened to the version, it's now 22 in the apworld which doesn't match the release number

Eijebong:
/taskcluster r+

🤷

PR #1677, opened 2026-03-31
Update KH:DDD

harmjan387:


Eijebong:
/taskcluster r+

Technically not a regression and it's been ok so far but someone smarter should have a look (I couldn't find a fix without rewriting create_items).

PR #1717, opened 2026-04-10
Update Hatsune Miku Project Diva Mega Mix+ to 1.4.4 UO2

RePod:
Allows Death Link for an indev in-game client until the next major apworld update that removes the Python client.

The Python client uses `deathLink` so there's no conflict despite `death_link` being forced to `True`.


Eijebong:
/taskcluster r+

That timeout cannot come from these changes, probably AP getting into swap hell

PR #1747, opened 2026-04-13
Update shadow_the_hedgehog

Eijebong:


Eijebong:
@choatix pinging you since there's fuzzer related code in the apworld :p Looks like you're missing some [indirect conditions](https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/apworld_dev_faq.md#i-learned-about-indirect-conditions-in-the-world-api-document-but-i-want-to-know-more-what-are-they-and-why-are-they-necessary).

choatix:
I've still a few logic issues in the world as well, but I'll attempt to get around to this before general release as well as the above other things which need to be improved. Looks to be a very consistent exchange for UT after a recent change a few versions back. I can also see that there's some issues internally to resolve with generation which I'll also endavour to fix.

PR #1748, opened 2026-04-13
Disable banjo tooie

Eijebong:


Eijebong:
Yeah ok I should've done this way earlier...

Eijebong:
/taskcluster r+

Who could've seen that one coming, turns out my manual testing was wildly inadequate. Disabling given the results but keeping the meta so future retries don't end up at 100% ignored

PR #1771, opened 2026-04-15
Added Metroid Prime to the apworld list

UltiNaruto:
0.5.3 now auto downloads libs so it doesn't have to include libs inside anymore

### DIFF
--- a/index/metroidprime.toml
+++ b/index/metroidprime.toml
@@ -3,5 +3,4 @@ home = "https://discord.com/channels/731205301247803413/1378465872917827614"
 default_url = "https://github.com/UltiNaruto/MetroidAPrime/releases/download/v{{version}}/metroidprime.apworld"
 
 [versions]
-"0.5.2" = {}
-"0.5.3" = {}
+"0.5.3-h1" = {}
--- a/index/metroidprime.toml
+++ b/index/metroidprime.toml
@@ -1,0 +1,7 @@
+name = "Metroid Prime"
+home = "https://discord.com/channels/731205301247803413/1378465872917827614"
+default_url = "https://github.com/UltiNaruto/MetroidAPrime/releases/download/v{{version}}/metroidprime.apworld"
+
+[versions]
+"0.5.2" = {}
+"0.5.3" = {}


Eijebong:
> 0.5.3 now auto downloads libs

Yeah unconditionally, on load. My infra doesn't load any apworld in any context where it can see any network. So unfortunately that's not not going to work.

UltiNaruto:
Main problem is having the patcher as a wheel. And new patcher is planned for v0.7.0. It won't use a wheel anymore when it's the new patcher.

Eijebong:
I don't understand why the patcher is required on generation? I would be fine if the client was the only part requiring it (since at that point it's on the user to trust it, you're downloading static files and it's no different than trusting a mod)

UltiNaruto:
So if I move the download lib part to client it's good?

Eijebong:
As long as it has no way to run during gen, I'm good with how it is yeah.

UltiNaruto:
I'll get a hotfix for it tomorrow then

UltiNaruto:
Made the changes, it should be good now

UltiNaruto:
With the upcoming v0.5.3-h2, I did a few fuzzer tests and got results ranging from 0% to 3% of failures with another game and 25% of failures when alone

Eijebong:
This has way too high of a failure rate. You definitely want to check those lambda capture issues. Might help a lot

UltiNaruto:
Got that after fixing the lambda capture bug
{"stats": {"total": 5000, "success": 3659, "failure": 0, "timeout": 0, "ignored": 1341}, "errors": {}}

Eijebong:
Ship it 🥳 (actually I hadn't seen the test failure but you'll also want to fix that)

UltiNaruto:
Which test failures?

Eijebong:
> Which test failures?

https://github.com/Eijebong/Archipelago-index/pull/1771#issuecomment-4290759492 This

UltiNaruto:
> https://github.com/Eijebong/Archipelago-index/pull/1771#issuecomment-4290759492 This

Just noticed it, I'm trying to figure it out

Eijebong:
/taskcluster test-fuzz

Just checking in case something else got missed

Eijebong:
Oh it was the same version, oops

UltiNaruto:
Fixed the tests :)
Now I need to release 0.5.4
EDIT: btw when I test fuzzer with tests fixed it's now failing more than earlier

PR #1772, opened 2026-04-15
add ff6wc

Koukin37:
This PR should add Final Fantasy 6 Worlds Collide to the Lobby


index/ff6wc.toml
```diff
@@ -0,0 +1,5 @@
+name = "Final Fantasy 6 Worlds Collide"
+home = "https://docs.google.com/spreadsheets/d/1iuzDTOAvdoNe8Ne8i461qGNucg5OuEoF-Ikqs8aUQZw/edit?gid=58422002#gid=58422002"
+
+[versions]
+"0.6.7" = { url = "https://bigmalletman.com/dl/ff6wc.apworld" }
```
Eijebong, line 5: That's not gonna work. That's a random versioned download URL which might change the content at any point.

Eijebong (commented):
I think I'll pass on this given the sheer amount of code to review, especially given the fact that there's so much binary in there. On top of the problem of the URL.

PR #1775, opened 2026-04-15
Added crabchampions.toml

Redstymp:
Initial version for crabcrampions 1.1.2
and update for 1.2.0

Suggestions to block bloated slots
```
# If on generates for each new pickup 1 check. Similar to dexsanity but worse since some upgrades can only be found if you received a different upgrade before.
pickup_checks: false
# When enabled, island completions and equipment runs are tracked per rank for ALL ranks up to Max Rank, not just the Required Rank.
# So depending on the run length minumum of 28 checks per weapon per rank configured.
extra_ranked_island_checks: false
# Can be turned on but still leads to alot of checks in the pool and a huge sphere 1 since they are all available
equipment_check_mode: disabled
```

### DIFF
--- a/index/crabchampions.toml
+++ b/index/crabchampions.toml
@@ -3,4 +3,4 @@ home = "https://discord.com/channels/731205301247803413/1295008123131658270"
 default_url = "https://github.com/Str8UpWHITE64/CrabChampionsAP/releases/download/{{version}}/crabchampions.apworld"
 
 [versions]
-"1.2.0" = {}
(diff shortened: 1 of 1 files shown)


Eijebong:
https://discord.com/channels/731205301247803413/1295008123131658270/1494893896633815051

Looks like some fixes are in the pipe for this so that's good

Eijebong:
/taskcluster r+

That's a lot better

PR #1777, opened 2026-04-16
Add Yellow Taxi Goes Vroom

soopercool101:
Adds Yellow Taxi Goes Vroom to the index. Note that this world has a lot of possible locations from its `coinsanity` option, which can be mitigated by the host via host.yaml settings. The default host settings allow for all of these locations to be enabled, but most to be forced filler. Feel free to reach out to me with any questions

### DIFF
--- a/index/yellowtaxi.txt
+++ b/index/yellowtaxi.txt
@@ -3,4 +3,4 @@ home = "https://discord.com/channels/731205301247803413/1265482079118950422"
 default_url = "https://github.com/soopercool101/Archipelago/releases/download/YTGV.v{{version}}/yellowtaxi.apworld"
 
 [versions]
-"0.3.1" = {}
+"0.3.2" = {}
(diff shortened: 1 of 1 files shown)


Eijebong:
`check-indirect-conditions` is a deal breaker. https://github.com/ArchipelagoMW/Archipelago/blob/main/docs/apworld_dev_faq.md#i-learned-about-indirect-conditions-in-the-world-api-document-but-i-want-to-know-more-what-are-they-and-why-are-they-necessary

soopercool101:
I have rule builder implemented on my branch, I'll release an update shortly with that which should address the issue. I'll also look into the determinism and UT failures

Eijebong:
FYI, determinism/ut are not blockers, just good to know.

soopercool101:
Rule builder fixed both the indirect condition and universal tracker issues, and the determinism fix was a one-line one. Should hopefully be all good now fuzzer-wise per local tests of the hooks. Let me know if anything further is necessary

Eijebong:
/taskcluster r+

Looks like legit restrictive starts

PR #1782, opened 2026-04-16
Enable Turnipboy

Gingerbit25:
Is the apworld actually bugged? Or it was disabled due to the fact that the player gets bk for long periods of time?
From my experience, that's normal


Eijebong:
1) This is wrong (wrong folder, not re-enabling)
2) Yes, it was disabled because receiving items in a specific order made people unable to receive more items https://discord.com/channels/731205301247803413/1297617351977734214/1473456014723514542

Gingerbit25:
Got it, sorry for the pr in the wrong folder I'm still learning

PR #1790, opened 2026-04-17
Add Hololive Treasure Mountain

SomeJakeGuy:


Eijebong:
/taskcluster test-fuzz

Honestly more for the tests than fuzzing as I'm expecting 100% failure due to that freetext

PR #1791, opened 2026-04-17
Add Nsmbw

SomeJakeGuy:


Mysteryem:
There is an entire template poptracker pack contained within the apworld, that only looks to have only been modified to add maps for Universal Tracker purposes. I think that could be reduced to just the maps file (and locations file once implemented).

`wii_code_tools` has been included in the apworld, but I don't see a license for this library in its repository, https://github.com/RoadrunnerWMC/wii-code-tools, which generally means All Rights Reserved, and would forbid redistribution without the author's explicit approval.

It seems like this apworld might still be a bit new and needing more testing/fixing before being ready for large multiworlds. 0.0.4 has some reports of some locations not sending, and the move rando option is a WIP according to the developer.

Yesterday

https://discord.com/channels/731205301247803413/1327187652864380948/1494947980703174758
> [developer] Locations: star coins, level completion, hint movies. Items: world unlocks and powerups (don't turn on move rando since it's WIP). Goal: beat Bowser which requires a number of worlds completed and star coin received.

https://discord.com/channels/731205301247803413/1327187652864380948/1494902315000401951
> [player] 1) Completing 9-1 did not send world4. Didn't test other levels of world 9.

https://discord.com/channels/731205301247803413/1327187652864380948/1495159754505584712
> [player] it doesnt send the "world1_level8_completed_level" check

> [developer] thats a bug, i've fixed it for the next version, I think your only current fix would be to do /send_location

PR #1792, opened 2026-04-17
Add Path of Exile

SomeJakeGuy:


Mysteryem:
LM looking kind of different.
image

----

Looks like they missed one direct use of `random` from when I'm guessing they replaced direct use of `random` with `world.random`, since the `import random` is still there in a few non-client files, but almost everything else is using `world.random` or `self.random`.


```python
def deprioritize_non_logic_gear(world: "PathOfExileWorld",
                                table: Dict[int, Items.ItemDict]) -> Dict[int, Items.ItemDict]:
    opt: PathOfExileOptions = world.options
    # If gear upgrades are disabled, don't try to deprioritize any gear, flask are already progressive.
    if opt.gear_upgrades.value == opt.gear_upgrades.option_all_gear_unlocked_at_start:
        return table
    required_categories = list()
    progression_main_gems = [gem for gem in Items.get_main_skill_gem_items(table) if
                             gem["classification"] == ItemClassification.progression]
    for gem in progression_main_gems:
        if gem.get("reqToUse"):
            required_categories.append(random.choice(gem.get("reqToUse")))
```

Also, `poe/poeClient/vendor/vendor_modules.zip` has a whole bunch of vendored libraries in it.

Eijebong:
That apworld is way too large... That's 10% of the rest of the index combined

PR #1794, opened 2026-04-18
Add Cursed Mario 64

MuratDev41:


Mysteryem:
This apworld is basically just the sm64ex world, but with a few extra locations defined and the maximum star count increased to 200. The apworld is undergoing rather rapid development at the moment, so 0.1.2 is already 3 versions behind (4 if you include a hotfix).

Going by the 0.1.5 hotfix notes
> Re-wrote ho filler is generated from based on Max Star Count to be based on remaining locations after progression items have been placed

I anticipate that older versions of the apworld will fail the `check-item-location-count`

MuratDev41:
@Eijebong if we pause the development until after chrism's sync would it be merged in or is something else required? Cause me and Incidious really want to join chrism's sync with this

Eijebong:
Can't gen its own template :p

MuratDev41:
oof

MuratDev41:
@Eijebong we fixed the fuzzer issue, this will be the last update for a while

Eijebong:
> Failure rate: 1.9%

This is still waaaay too high

MuratDev41:
What is "no restrective starts", and how can i check it with the fuzzer?

kingdragonite55:
Using https://github.com/Eijebong/empty-apworld. Install empty.apworld, create a folder with the Empty yaml in it, then run the fuzzer with `--with-static-worlds <folder>`

PR #1798, opened 2026-04-19
Add version 1.2.2 to deltarune.toml

Tenebrosful:
That's a pre-release with game mod port to a new framework for ch1 and chapter select for now but should be all fine 👍

### DIFF
--- a/index/deltarune.toml
+++ b/index/deltarune.toml
@@ -12,4 +12,4 @@ default_url = "https://github.com/theemeraldsword85/DELTARUNEAP/releases/downloa
 "1.1.4" = {}
 "1.2.0" = {}
 "1.2.1" = {}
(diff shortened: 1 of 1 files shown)


Tenebrosful:
well guess i will leave it as 1.2.2 for the tag

Eijebong:
/taskcluster diff

Looks like the artifact got overriden (🫠) and that diff looks like the apworld was missing files?

Eijebong:
Nope. Still looking wrong 🤷‍♂️ I'll have a look tomorrow 

Edit: nvm, damn caches. It was a buggy artifact

Tenebrosful:
I think the release file was wrong but after some reflexions in think we will wait the full porting since we marked it as pre-release

PR #1799, opened 2026-04-19
Update lego_star_wars_tcs.toml to v1.4.0

thesunisjustfine:
updated to 1.4.0

im not the dev of this apworld; i just want to play this in the community ap (no other people have posted a yaml for this game yet)

this is also my first pull request so if it needs something changed please let me know!


Mysteryem:
Note that this will fail the UT hook because the UT hook has a bug where it does not pass slot_data through json conversion before feeding it into the UT passthrough: https://github.com/FarisTheAncient/Archipelago/pull/130

In normal operation, core AP turns lists in slot_data into tuples before writing them into the multidata. When hosting an ap server, the slot_data gets dumped to json, sent to the client and then loaded from json on the client, which results in all tuples from the multidata's slot_data being converted to lists when received by the client. So this issue is only present in the UT hook, and not at actual runtime with UT.

PR #1800, opened 2026-04-19
Add System Shock 2

Mute-255:


Mysteryem:
I suspect this client is going to spam the server with `LocationChecks` messages (once every 0.3s) because it is not updating its `ctx.sent_items` due to using the wrong method.

In `ss2.client.loc_watcher()`:
```python
                    if new_locs:
                        locs.extend(new_locs)
                        ctx.sent_items.union(new_locs)
```
`ctx.sent_items.union(new_locs)` creates a new set that is the union of `ctx.sent_items` and `new_locs`. This code should likely have been `ctx.sent_items.update(new_locs)` because the current code effectively does nothing.

I would strongly recommend that all `CommonClient` based clients use the `CommonContext.check_locations()` helper async function to send `LocationChecks` instead of sending `LocationChecks` messages manually, because it filters out the already checked locations and would have preventing spamming the server even with the above bug.

Replace
`asyncio.create_task(ctx.send_msgs([{"cmd": "LocationChecks", "locations": locs}]))`
with
`Utils.async_start(ctx.check_locations(locs))`

The use of `asyncio.create_task()` in the client code is also a bug because no reference is kept to the created task, which means it could get garbage collected before being run (the event loop only keeps weak references to tasks, see the **Important** section in https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

Locally hosted ap servers have a `--log_network` argument that can be enabled by developers to check that their client is not spamming the server, which I recommend developers use occasionally to make sure their clients are working as intended.

Mysteryem:
Some of the logic in the apworld looks like it has a lot of room for improvement, which is not necessarily blocking, but could be quite slow for how simple the logic should be.

`cyb_mod_count()` is called by a lot of rules and should really be replaced with `collect()` + `remove()` overrides on the world class that collect/remove a count of a fake "Cyber Modules" item equivalent to the count of the item. Though `"Naturally Able OSUpgrade"` would need a little extra work in `collect()`/`remove()` to only give/remove the +20 for the first `"Naturally Able OSUpgrade"` collected/removed.
```python
    def cyb_mod_count(self, state) -> int:
        curcybmodamount = 0
        curcybmodamount += state.count("2 Cyber Modules", self.player) * 2
        curcybmodamount += state.count("3 Cyber Modules", self.player) * 3
        curcybmodamount += state.count("4 Cyber Modules", self.player) * 4
        curcybmodamount += state.count("5 Cyber Modules", self.player) * 5
        curcybmodamount += state.count("6 Cyber Modules", self.player) * 6
        curcybmodamount += state.count("7 Cyber Modules", self.player) * 7
        curcybmodamount += state.count("8 Cyber Modules", self.player) * 8
        curcybmodamount += state.count("10 Cyber Modules", self.player) * 10
        curcybmodamount += state.count("13 Cyber Modules", self.player) * 13
        curcybmodamount += state.count("14 Cyber Modules", self.player) * 14
        curcybmodamount += state.count("15 Cyber Modules", self.player) * 15
        curcybmodamount += state.count("16 Cyber Modules", self.player) * 16
        curcybmodamount += state.count("20 Cyber Modules", self.player) * 20
        curcybmodamount += state.count("25 Cyber Modules", self.player) * 25
        curcybmodamount += state.count("30 Cyber Modules", self.player) * 30
        if state.has("Naturally Able OSUpgrade", self.player):
            curcybmodamount += 20
        return curcybmodamount
```

The main place that uses `cyb_mod_count()` is as the last argument to `upgrade_or_cybmod()`, which doesn't even use the last argument if the `include_stats_skills_psi` option is enabled.

```python
    def upgrade_or_cybmod(self, state, item, amount, cybmodamount, curcybmodamount) -> bool:
        if self.options.include_stats_skills_psi:
            return state.has(item, self.player, amount)
        else:
            return cybmodamount <= curcybmodamount
```

It is recommended to set different rules depending on options, because rules may be checked 1000 to 10000 times or more during generation. Doing as much work prior to calling the rules as possible generally means better generation performance.

Overall, my recommendation would be getting the world's logic swapped over to using Rule Builder, since that avoids most of the logic performance problems here (besides the `collect()`+`remove()` overrides). Using Rule Builder would also get rid of some of the other wasteful calculations in rules like doing `(a*2)-2` where `a` is a static value. If used correctly, everything in Rule Builder that is static is calculated up-front, including checking against options, and avoids lambda capture issues.

Mysteryem:
Unfortunately, it seems the current developer is stepping away from the Archipelago implementation as of yesterday, so who knows when there will be updates to the System Shock 2 AP implementation.
https://discord.com/channels/731205301247803413/1340419754200010915/1495651695873491025

Eijebong:
That is very unfortunate :/ I'll mark this as blocked because that kind of client bug spamming the server is really bad for large syncs (the server is busy enough as it is)

Mute-255:
Unfortunate but understandable, thanks for taking the time to look at it.

PR #1802, opened 2026-04-19
Update Manual: Titanfall 2 to v0.16

niyrme:


niyrme:
hold on, gotta yoink that gitattributes

PR #1803, opened 2026-04-20
Update bfm.toml to v0.5.0

AegeusEvander:
Adds open world support and bug fixes


Mysteryem:
I noticed the client has changed to using `ctx.check_locations()`, which is a good change if it happened to be spamming the server in older versions.

The logic performance for this world would probably significantly benefit from being converted to Rule Builder, so that the logic stops looking up the correct name of every item used from within rules, checking against options in rules, or creating new sets to pass into `has_all()`/`has_any()` just to immediately throw those sets away, since Rule Builder would simplify doing all of that upfront, instead of doing all that within the rules themselves.

/taskcluster test-fuzz

index/bfm.toml
```diff
@@ -3,4 +3,4 @@ home = "https://discord.com/channels/731205301247803413/1291453297370337351"
 default_url = "https://github.com/AegeusEvander/Brave-Fencer-Musashi-AP-World/releases/download/v{{version}}/bfm.apworld"
 
 [versions]
-"0.4.2" = {}
+"0.5.0" = {}
```
Mysteryem, line 7: Usually, the new version is added alongside the old version. Is there some reason the old version needs to be removed?

```suggestion
"0.4.2" = {}
"0.5.0" = {}
```

AegeusEvander: Nah old version is fine too

PR #1812, opened 2026-04-23
Update Black Ops 3 - Zombies to 0.5.1

colin969:


colin969:
Lambda capture issues should be fixed now, please run fuzz again

Microslops api for... editing a pull request title has been broken for weeks now, so I still can't update that one.

PR #1831, opened 2026-04-23
Pretty Derby YMAL Update

QuindoHL:
Added YMAL configuration options so the apworld's size can scale to the user specific account unlocks.


Eijebong:
URL is wrong

QuindoHL:
/taskcluster test-fuzz

Not sure if only you can do this but if it saves ya time and mental energy Ill try to get a head start on it.

QuindoHL:
FYI, the pre-exsisting AP is compatible with 6.7 if you want to pause this PR until after this weekend.

